PR #7, opened 2017-03-16
Renaming: use original titles

lex111:
For example:
- Frontend -> Front-end
- PHP-7 -> PHP 7
- Node JS -> Node.js
- NPM -> npm

And so on, in general, the names have been renamed in such a way as they are used officially.

Maybe it does not really matter, but still I want to use the correct technology names, so that there is no confusion


kamranahmedse:
Thanks

PR #35, opened 2017-03-20
Add loopback.io to javascript section and Zend to PHP 7 section

noroot:
Add loopback.io to javascript section and Zend to PHP 7 section


kamranahmedse:
Thank you for taking the time to create PR. I am well aware of these and I intentionally left these out as the repository focuses on what's "hot" in 2017 and I don't think we should include each and everything.

Thanks though.

PR #45, opened 2017-03-21
New relic

DabHands:
I added New Relic to the list of monitoring tools. Is there a better way to monitor your infrastructure?


diego-vieira:
@DabHands what about sentry?

DabHands:
@diego-vieira I have never used Sentry but a quick glance at their features reveals that it will be a great tool. The truth is there are a lot of tools out there. Some will make the list some won't.

ascendbruce:
@diego-vieira @DabHands I guess New Relic, Skylight, Scout are similar tools (performance monitoring). And Sentry, Bugsnag, AirBrake are similar (exception notifier).

diego-vieira:
@ascendbruce indeed, that makes more sense.

DabHands:
@ascendbruce you are right. But the heading is 'Monitoring and Alerting' not 'Performance monitoring and Exception notifier'. So I believe the heading 'Monitoring and Alerting' captures everything. Where @diego-vieira and I could have gotten it wrong is to think New Relic and Sentry perform the same function.

PR #48, opened 2017-03-22
c# expanded .net and .net core

ebattulga:
expanding .net to asp.net and asp.net core


iCodeWebApps:
It would be better if you followed the established pattern at defining a language.
For example:

C#
1. Framework
   Asp.net
   Asp.net MVC
   Asp.net Core
2. Package Manager
   NuGet Package Manager
3. Testing
   MTM
   NUnit

kamranahmedse:
Hi @ebattulga, 

Thank you for the PR but I wouldn't be able to merge this in. The graph is already too much convoluted and it is becoming difficult to even read. So let's keep .NET to the leaf for now, till we figure out how to fit in everything.

Thank you.

PR #63, opened 2017-03-24
Added Video Discussing Roadmap in Detail

rpz-harode:
Video discusses the current trends of web-development technologies and answers an important question - What to be learned in 2017 .


kamranahmedse:
Hi @rpz-harode, thank you for the PR but the video seems to be just the plain translation of what is in image and is missing a lot of stuff that we have in here. So, I think it is fine. Thanks though.

PR #69, opened 2017-03-24
Added several databases

dverstee:


kamranahmedse:
Thank you. I know there are several which are missing but it is intentionally kept simple and focuses on the most demanded ones. Thanks though.

PR #71, opened 2017-03-24
Improved DevOps part

Applemann:
I was improved server section, added backups, virtualisation, docker ekosystem, fix in monitoring, shell and operating system.. :)


kamranahmedse:
Hey Martin,

While I agree with what you added but it intentionally does not go deep into each of those technologies to avoid the convolution; so unfortunately I have to close this one but thank you for taking the time to create the PR.

Applemann:
@kamranahmedse  Ok so what is a lot deep? I can remove something.. (Docker environment, virtualization..?)

PR #74, opened 2017-03-25
Added recommended editor/IDEs (Front-end) & python libraries (Backend)

repositivator:
Added some recommended editors & IDEs (for frontend)
- Sublime Text
- Visual Studio Code
- Brackets
- Web Storm
- Atom

Added some famous python libraries.
- **Basic data analysis** : Tensorflow, Pandas, NumPy, SciPy, Scikit-learn 
- **Web crawling** : Beautiful Soup, Scrapy

Thank you very much for making neat & nice roadmap. :)


kamranahmedse:
While I agree with the editors but I would like to put them in simple textual form and for the libraries; we don't have a place for them (it is just a hight level map).

PR #81, opened 2017-03-26
Make JSON code readable and easier to maintain

jbaron-mx:
This PR does not introduce any new technology to the charts, it just _prettify_ all current JSON files, the benefits are:

- It makes code **easier to read**, because all the code is formatted and indented properly.
- It makes code **easier to maintain**, because you can no longer have diffs where it's so difficult to see what changed.

I've tested on Balsamiq 3.5.8 and still works as expected.


kamranahmedse:
Thanks but I don't think anybody is going to open this JSON file and modify it directly as it is meant to be used in Balsamiq only which is fine without formatting.

PR #115, opened 2017-04-02
Update README.md

march93:
Small fix to README


kamranahmedse:
Hi @march93 thank you for the PR but "do suggest" is fine.

PR #118, opened 2017-04-05
Added Windows OS

SatyaSudheer:
Added Windows Option in the operating systems.


esuiru:
Agree, Windows Server, especially in devops where Azure is an option.

kamranahmedse:
Although windows is an option but if you look at it linux/unix is going to be always preferred over it; so I'll have to pass. But thanks for taking the time to create PR.

Issue #149, opened 2017-04-26
[suggestion] Add 'nose' to Python testing tools 


epogrebnyak:
It would be nice to see unittest, pytest and nose listed together. As some example: these three tools are supported by TravisCI, a popular continios integration tool.

splinterific:
I put through a PR on this at https://github.com/kamranahmedse/developer-roadmap/pull/155 :)

Kwpolska:
http://nose.readthedocs.io/en/latest/

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

Issue #143, opened 2017-04-24
Add python framework asynchronous


monobot:
You should add on Python > Framework > asynchronous > sanic https://github.com/channelcat/sanic

splinterific:
I've added this in a PR :)

kamranahmedse:
Added. Thanks

PR #155, opened 2017-04-29
added nose to Python -> Testing

splinterific:
bulk adding changes for #149 and #143 so you don't get compile issues merging two branches on the image


splinterific:
Looking at the PR, you'd rather images be included. One point however is that all the images together will vary based on each PR. If you want to set a deadline and pull in each PR, I can then re-add an image that contains all the updates. Otherwise the images will be staggered

kamranahmedse:
Added. Thanks @splinterific

PR #277, opened 2017-08-30
Added ASP.NET Core to .NET (C#) section

vnbrs:
ASP.NET Core is a .NET web framework to build scalable, modern and performatic web applications. I think it should be included in the .NET roadmap section.


Jogai:
Especially important for new devs or those coming from another framework.

PR #281, opened 2017-09-15
Chinese version of "Web Developer Roadmap in 2017"

z-jingjie:
This is a Chinese translation of "Web Developer Roadmap in 2017" (lastest version matched), please review it & see if it's ok for merging or linking, thanks.


kamranahmedse:
Hey there,

Thanks for opening the PR but unfortunately we don't have any plans to add translations as it would be really difficult to keep all the maps in sync.

But thanks though.

PR #304, opened 2018-01-05
Added elm to frontend map

inammathe:
https://i.imgur.com/PKa5H5V.png


lostdesign:
The resolution of that roadmap is quite low. Should be fixed imo.

kamranahmedse:
Thank you for opening the issue. While I agree with that, we don't want to convolute the roadmap by mentioning every available option and are just focusing on what is the most relevant item that you would pick today and mentioned item doesn't conform to that. 

This has also been [mentioned in the contribution docs](https://github.com/kamranahmedse/developer-roadmap/blob/master/contributing.md#guidelines). Thank your for your suggestion though 🙏

PR #328, opened 2018-01-11
Added a list of Linux & Unix server distributions, and Emacs as text editor

da-edra:
I'm making this PR to include some Linux and Unix distributions that are ideal for web servers as not all distributions are suitable to be used in a production server and it can be hard to choose one if you're new to Linux.

I've also added Emacs as a text editor since it has some amazingly powerful features and it's the closest contendant to Vim.


thorerik:
Honestly, I don't think MacOS Server has any place on that map, there are very few use cases for MacOS server, and those are usually limited to the Mac equivalent of a local AD/Windows environment, just for macs instead (Patch management, intranet, backup, etc.)

If you want more than 1 UNIX, maybe DragonflyBSD or NetBSD instead, as they are far more suited for a DevOps kind of environment.

As for the Linux-es, maybe highlight CentOS and Ubuntu as well, since RHEL is license bound and there's no easy way for someone learning to pick one up for learning purposes (though there are dev licenses available, though they aren't obvious how to obtain), since CentOS is going to teach you all the necessary things for using RHEL.

da-edra:
Good points, @thorerik I'll make a new commit with the changes. 😄 
I think I'll remove MacOS Server and add OpenBSD (2nd most popular BSD) and NetBSD since DragonFlyBSD is a fork of FreeBSD.

da-edra:
I've added the new changes and updated the image, it looks better now. :)

michealwisdom19850707:
I need best server

kamranahmedse:
👍

PR #348, opened 2018-01-17
added google polymer.js as webcomponent library.

ankitsorathiya:
Web component development is great and we have google polymer which helps to build web component.  I personally have tried and built web products using google polymer.
I think it would add good values if we can include that in to front end road map.


kamranahmedse:

Issue #296, opened 2017-12-09
Improving the Angular roadmap


envil:
1. Separate Angular and Angular JS
2. Add ngrx/store (redux like solution inspired by rxjs) to Angular

kamranahmedse:
Hey, thanks for reporting. 

Yes I do have that in mind but rather planning to do it once and for all in twenty days on `January 1st, 2018` :)

kamranahmedse:
Pushing in a few. Thanks for opening.

envil:
@kamranahmedse kam
Thank you for updating. I think it's worth to include ngrx (state management for Angular).

Issue #26, opened 2017-03-19
JavaScript Frameworks


MikhailTatsky:
Please add in Front-end, JavaScript Frameworks
ExtJS, EmberJS

bennyn:
I am also missing [Polymer](https://www.polymer-project.org/).

ChouUn:
pls add in front-end: rxjs, immutable.js, backbone, lodash / ramda / underscore.

link-alex:
Is it really suitable to put all the js frameworks on the chart? like, ALL of them? And all the cool libraries should be there also in this case?
I believe a separate chart for this topic is required then..
What do you think guys?

arpit1997:
@ChouUn I agree with @link-alex Its not suitable to pull off every cool javascript library in one chart there is like hundereds. We can try making a list by popularity or category.

carlosgalveias:
+1 for EmberJS, its a very popular framework used by Netflix, LinkedIn, Apple Music and its awesome!!

se-panfilov:
In terms of trending frameworks, I'd say Vue.js and Aurelia

deerua:
angularJs and angular (2.x > ) now separate stuff 
and angular can use redux, rxjs (core stuff of angular)

kamranahmedse:
Ember JS agreed (pushing it in a few minutes). But as @link-alex said, if we keep adding each and everything it will easily become convoluted. Thank you.

luchillo17:
@deerua Is right, angular 2 can now use Redux through `@ngrx/store` npm module which uses `Rxjs` to make redux Observable based, however i don't think Ng1 & Ng2 should be considered separated in the graph, Ng2 is indeed a full rewrite of the framework, but that's just a peculiarity between the 2 versions that shouldn't be included in the graph.

mcMickJuice:
Why not separate out flue/redux into a "state management" section?  By principal they're UI library agnostic and can be used with nearly any frontend framework.

PR #353, opened 2018-01-18
Add ngrx to Angular

jansivans:
Added ngrx to Angular as requested here: #296, #26


kamranahmedse:
Frontend has be redrawn and it is there now. Thankk you for the PR though.

PR #394, opened 2018-02-13
Added Cypress

mihhu:


kamranahmedse:
There have been changes made to that section which makes this issue outdated. Thank you for reporting though 🙏

PR #404, opened 2018-02-18
Add ASP.NET Core to C#.NET Path

NageshAndani:
Modified Backend project file to add ASP.NET Core and ASP.NET Framework 4.7 to C#.NET path


NageshAndani:
Expecting a note on what's wrong in this commit. Is there a better way it could be updated?

MartinZikmund:
I think ASP. NET Core should be included. ASP. NET is now rather legacy and Core is significantly different

PR #425, opened 2018-03-10
Bengali Translations initiated

zonayedpca:
I want to translate it into my own language so that our people can understand it more easily. Please let me know if it is allowed or not. Thanks :)


kamranahmedse:
Hey, thanks for the PR. I would rather suggest you to do it in the fork and then share the link with me, I will put it in the repository :)

zonayedpca:
Thanks bro... Can I use these images in my blog post?

raunakhajela:
What images?

kamranahmedse:
Yes, as long as you refer back to the repository.

PR #501, opened 2018-10-08
Added some resource websites

akshatah33:


voyula:
-1

Resources is already unlimited.

kamranahmedse:
Thank you for the PR. We are already [working on something](https://roadmap.sh). I would rather wait for that before accepting any PRs on the resources. Thank you!

Issue #476, opened 2018-07-25
Image transparency


2E0PGS:
Greetings,

For us dark theme users who appreciate transparency in images could you please update the rest of the images to be transparrent aswell.

The following already have transparrent backgrounds:

* https://github.com/kamranahmedse/developer-roadmap/blob/master/images/backend.png
* https://github.com/kamranahmedse/developer-roadmap/blob/master/images/devops.png
* https://github.com/kamranahmedse/developer-roadmap/blob/master/images/split.png

The following need updating to be transparrent background. I could do this myself but if you have the source project files for those images then it would save me trying to use magic wand tool on it and getting inaccurate clipping of images:

* https://github.com/kamranahmedse/developer-roadmap/blob/master/images/disclaimer.png
* https://github.com/kamranahmedse/developer-roadmap/blob/master/images/frontend-v2.png

Example of how I detected this issue:

![image](https://user-images.githubusercontent.com/15659301/43194363-13fd3c20-8ffa-11e8-8b13-34672d62c227.png)

Notice how the latter image fits in well with my dark CSS theme since it has a transparrent background.

1337M0nst3r:
Working on it!

kamranahmedse:
Thank you for reporting. We have fixed the transparency of the roadmaps in the repository. If you are still looking for the transparent backgrounds, you can find the images at [roadmap.sh](https://roadmap.sh).

PR #503, opened 2018-10-11
add transparency to disclaimer.png and frontend-v2.png

1337M0nst3r:
Issue #476


Jogai:
I'm afraid this will clash with https://github.com/kamranahmedse/developer-roadmap/issues/535

I'm using github dark script which is a bit blacker and then the black text in the transparent background is hardly legible.

kamranahmedse:
We have removed the transparency on all the images so it is not needed any more. Thank you!

PR #505, opened 2018-10-13
him -> them

jfdnc:
matching with the neutral term "user"


kamranahmedse:
There have been changes made to that section which makes this issue outdated. Thank you for reporting though 🙏

PR #519, opened 2018-11-29
fix the cut off names teamcity, LXC, and puppet

johnoct:
The names were cutoff so I updated the spacing so its clear.


eko24ive:
pinging @kamranahmedse to merge/use this one, because I almost cloned this repo to fix this until I figured out that there is already might some fix

kamranahmedse:
Thanks

PR #551, opened 2019-01-21
Typo fix

bobbykjack:
Mainly to replace "push push" with "push", but then to clean up for alignment too. The nature of the json file (all one line) makes the diff awkward, I'm afraid, but I've tested this in Balsamiq and it looks good.


Ronkiro:
This seems ok, but i think you forgot to update the image (images/frontend.png)

kamranahmedse:

PR #587, opened 2019-05-15
Fix word duplication in the CSS Architecture comment box. Removed 'push'.

ericjeker:
Fixed that duplicated word.

image


kamranahmedse:
Thanks

PR #593, opened 2019-05-29
Update ELK Stack to Elastic stack within devops-map.json

fhnbg:
As ELK Stack is renamed to Elastic Stack i updated devops-map.json


kamranahmedse:

PR #603, opened 2019-07-08
Arabic translation

shu3aybbadran:
Hi,
I translated your roadmap to Arabic,
but not sure what the right way to merge this


kamranahmedse:
Thank you. We can't merge it due to the nature of these projects. However, we are working on [the website](https://roadmap.sh) which will allow easier maintenance and introducing translations. I will reach out to you for that :)

PR #605, opened 2019-07-09
Remove JSCS (deprecated) and add StandardJS linter

harobed:
Remove JSCS ([deprecated](https://www.npmjs.com/package/jscs)) and add [StandardJS](https://standardjs.com/) linter.


kamranahmedse:
Thanks

Issue #625, opened 2019-10-24
Sass, not SASS


TheJaredWilcurt:
Sass has officially never been all caps. Similar to how npm is never NPM.

SuperQ:
Usually it's ["SaaS"](https://en.wikipedia.org/wiki/Software_as_a_service).

TheJaredWilcurt:
@SuperQ This is not related to "Software as a Service".

[https:///Sass-Lang.com](https://sass-lang.com)

Ronkiro:
@kamranahmedse I created a PR to solve this issue.
Please consider checking it :)

kamranahmedse:
This has been fixed and will be published in the next release. Thank you!

PR #626, opened 2019-10-29
Changed SASS -> Sass

Ronkiro:
According to issue #625, SASS shouldn't be referenced as SASS, but as Sass.


kamranahmedse:
This has been fixed and will be published in the next release. Thank you!

PR #1259, opened 2022-05-11
Add Status Buttons

Magdy98:
Add 3 Buttons display percentage of learning .


kamranahmedse:
This does seem interesting but there are no plans to add it to the project. Thank you for your contribution though 🙏

PR #1416, opened 2022-08-06
Simple understanding of Semantics

wisdak:
I guess this might be helpful


kamranahmedse:
Thank you for opening the PR but I don't think this has anything to do with "Semantic HTML".

PR #1492, opened 2022-08-13
Create readme-pt.md

henriquesebastiao:
Creating a readme file in Portuguese.


kamranahmedse:
Thank you for your contribution. We do have localization planned for the website but it is not in the near term.

PR #1559, opened 2022-08-19
Update 102-transactions.md

musadabra:
This is a basic explanation of the Transaction, if this doesn't convince you, I would love to give a more detailed explanation.

### DIFF
--- a/content/roadmaps/101-backend/content/108-more-about-databases/102-transactions.md
+++ b/content/roadmaps/101-backend/content/108-more-about-databases/102-transactions.md
@@ -1,1 +1,3 @@
 # Transactions
+
+A transaction generally represents any change in a database, this symbolizes a unit of work performed within a database management system against a database, and treated in a coherent and reliable way independent of other transactions. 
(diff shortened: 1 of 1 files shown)


content/roadmaps/101-backend/content/108-more-about-databases/102-transactions.md
```diff
@@ -1 +1,3 @@
-# Transactions
\ No newline at end of file
+# Transactions
+
+A transaction generally represents any change in a database, this symbolizes a unit of work performed within a database management system against a database, and treated in a coherent and reliable way independent of other transactions.
```
kamranahmedse, line 3:
```suggestion
# Transactions

In short, a database transaction is a sequence of multiple operations performed on a database, and all served as a single logical unit of work — taking place wholly or not at all. In other words, there's never a case where only half of the operations are performed and the results saved.

<BadgeLink colorScheme='yellow' badgeText='Read' href='https://fauna.com/blog/database-transaction'>What are Transactions?</BadgeLink>
```

PR #1583, opened 2022-08-23
Add Content into Backend-Developer-Roadmap

madilekci:
Add content into Backend-developer-roadmap -> Learn about APIs -> Authentication


kamranahmedse:
Thanks for the contribution @madilekci. I have [made some changes](https://github.com/kamranahmedse/developer-roadmap/pull/1583/commits/277bea3de23a0745a673b421153f618be085575d), please have a look. There were a few dangling links and also we shouldn't be writing too many details in the description.

madilekci-avena:
Yes you are right, thanks for the edit.

PR #1618, opened 2022-08-27
Update  108-javascript-expressions-and-operators

h-s04:
- Update JavaScript expressions and operators

### DIFF
--- a/content/roadmaps/106-javascript/content/108-javascript-expressions-and-operators/readme.md
+++ b/content/roadmaps/106-javascript/content/108-javascript-expressions-and-operators/readme.md
@@ -1,1 +1,13 @@
-# Javascript expressions and operators
+# JavaScript expressions and operators
+
+At a high level, an expression is a valid unit of code that resolves to a value. 
+There are two types of expressions: those that have side effects (such as assigning values) and those that purely evaluate.
+
+The expression `x = 7` is an example of the first type. This expression uses the `=` operator to assign the value seven to the variable x. The expression itself evaluates to 7.
+
+The expression `3 + 4` is an example of the second type. This expression uses the `+` operator to add `3` and `4` together and produces a value, `7`. However, if it's not eventually part of a bigger construct (for example, a variable declaration like const `z = 3 + 4`), its result will be immediately discarded `—` this is usually a programmer mistake because the evaluation doesn't produce any effects.
+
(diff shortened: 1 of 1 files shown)


content/roadmaps/106-javascript/content/108-javascript-expressions-and-operators/readme.md
```diff
@@ -1 +1,13 @@
...
+There are two types of expressions: those that have side effects (such as assigning values) and those that purely evaluate.
+
+The expression `x = 7` is an example of the first type. This expression uses the `=` operator to assign the value seven to the variable x. The expression itself evaluates to 7.
+
+The expression `3 + 4` is an example of the second type. This expression uses the `+` operator to add `3` and `4` together and produces a value, `7`. However, if it's not eventually part of a bigger construct (for example, a variable declaration like const `z = 3 + 4`), its result will be immediately discarded `—` this is usually a programmer mistake because the evaluation doesn't produce any effects.
+
+As the examples above also illustrate, all complex expressions are joined by operators, such as `=` and `+`.
+
+<ResourceGroupTitle>Free Content</ResourceGroupTitle>
+<BadgeLink colorScheme='blue' badgeText='Read' href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators'>Expressions and operators</BadgeLink>
```
kamranahmedse, line 13:
```suggestion
# Expressions and Operators

At a high level, an expression is a valid unit of code that resolves to a value. There are two types of expressions: those that have side effects (such as assigning values) and those that purely evaluate. The expression `x = 7` is an example of the first type. This expression uses the `=` operator to assign the value seven to the variable x. The expression itself evaluates to 7. The expression `3 + 4` is an example of the second type. This expression uses the `+` operator to add `3` and `4` together and produces a value, `7`. However, if it's not eventually part of a bigger construct (for example, a variable declaration like const `z = 3 + 4`), its result will be immediately discarded `—` this is usually a programmer mistake because the evaluation doesn't produce any effects. As the examples above also illustrate, all complex expressions are joined by operators, such as `=` and `+`.

<ResourceGroupTitle>Free Content</ResourceGroupTitle>
<BadgeLink colorScheme='blue' badgeText='Read' href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators'>Expressions and operators</BadgeLink>
```

PR #1622, opened 2022-08-27
Node.js content

another-m00:
I started to make the node.js content in this fork. It is strongly in Work in progress stage since only 2 node is done, and those still need to be reviewed.
I still wanted to make a pull request to show the existing progress.
I will probably issue some changes in the map structure itself, since debugging for example is a pretty important part of coding and there are no reliable resources online for debugging in node.js

Sooner or later I will make this into a pullable state 😃


another-m00:
Currently this is in regular markdown, i didnt use any of the react components

kamranahmedse:
Thanks, @another-m00. I force-pushed to the repo. Can you please create a new fork and resubmit the PR.

Also, I went through the markdown files -- we don't have to give a thorough explanation for each of the node topics. Please just add 2 to 3 lines giving a brief introduction about what the given item is and focus on adding links.

For example, here is a sample node from one of the other roadmaps 

image

As you can see, we did not delve into the details of the topic, just a few introductory lines followed by the resources.

another-m00:
It's kind of a pity that these are just stubs that dont even give an overview of the application itself, but okay, sure

PR #1675, opened 2022-09-03
typo in title and file name

hchiam:
I wish I knew about a resource like this when I first started! I'm thinking of helping fill in some of the unfilled content, but if I get around to it, I'll do it in separate PRs.


hchiam (commented):
Ready for review.

kamranahmedse:
Thank you for your contribution Howard, but this will break the roadmap interactivity. There are multiple places where this needs to be changed: in the roadmap, the paths file, and the file. Let me do it separately.

PR #1688, opened 2022-09-04
Added description and resources to Lexical Scoping - JavaScript Roadmap

codeguage-code:


kamranahmedse (commented):
Thank you for your contributions. Apart from all the links from your website, can you please add 2 to 3 links from different websites as well :)

codeguage-code:
Yeah sure. Even I felt that way, and that's why some of the latest commits refer to MDN resources as well. But yeah, I'll surely try.

PR #1712, opened 2022-09-07
Update readme.md

Muzzammmill:
Hi.
I thought it would be good to have at least an ES6 javascript course


kamranahmedse:
Thanks Muzammil, but there are already more than enough courses in the list, we don't want a convoluted list so let's not add it.

Muzzammmill:
Ok, thanks for the time.

PR #1722, opened 2022-09-10
Update 105-what-is-hosting.md

yashdev9274:
edit the definition of hosting .


kamranahmedse:
Thanks Yash but the existing definition is already good enough.

yashdev9274:
no worry , I'll update other content of it.

PR #1739, opened 2022-09-12
Update readme.md

skieurfoufou:


kamranahmedse:
Thanks but "excerpt" is the valid word here.

PR #1814, opened 2022-09-20
Update 101-value-comparison-operators.md

ivaylokarafeizov:
Added information about value comparison operators: ==, === and Object.is(). Also added links with more free content to read.

### DIFF
--- a/content/roadmaps/106-javascript/content/105-javascript-equality-comparisons/101-value-comparison-operators.md
+++ b/content/roadmaps/106-javascript/content/105-javascript-equality-comparisons/101-value-comparison-operators.md
@@ -1,1 +1,11 @@
 # Value comparison operators
+
+In javascript the == operator does the type conversion of the operands before comparison, whereas the === operator compares the values as well as the data types of the operands. The Object.is() method determines whether two values are the same value: Object.is(value1, value2).  
+
+Object.is() is not equivalent to the == operator. The == operator applies various coercions to both sides (if they are not the same type) before testing for equality (resulting in such behavior as "" == false being true), but Object.is() doesn't coerce either value.
+
+Object.is() is also not equivalent to the === operator. The only difference between Object.is() and === is in their treatment of signed zeros and NaN values. The === operator (and the == operator) treats the number values -0 and +0 as equal, but treats NaN as not equal to each other.
+
+<ResourceGroupTitle>Free Content</ResourceGroupTitle>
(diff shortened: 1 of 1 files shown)


content/roadmaps/106-javascript/content/105-javascript-equality-comparisons/101-value-comparison-operators.md
```diff
@@ -1 +1,11 @@
...
+
+Object.is() is not equivalent to the == operator. The == operator applies various coercions to both sides (if they are not the same type) before testing for equality (resulting in such behavior as "" == false being true), but Object.is() doesn't coerce either value.
+
+Object.is() is also not equivalent to the === operator. The only difference between Object.is() and === is in their treatment of signed zeros and NaN values. The === operator (and the == operator) treats the number values -0 and +0 as equal, but treats NaN as not equal to each other.
+
+<ResourceGroupTitle>Free Content</ResourceGroupTitle>
+<BadgeLink colorScheme='yellow' badgeText='Read' href='https://www.scaler.com/topics/javascript/difference-between-double-equals-and-triple-equals-in-javascript/'>The Difference Between == and === in Javascript</BadgeLink>
+<BadgeLink colorScheme='yellow' badgeText='Read' href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is'>More Info about Object.is()</BadgeLink>
```
kamranahmedse, line 11:
```suggestion
# Value Comparison Operators

In javascript, the `==` operator does the type conversion of the operands before comparison, whereas the === operator compares the values and the data types of the operands. The `Object.is()` method determines whether two values are the same value: `Object.is(value1, value2)`.

`Object.is()` is not equivalent to the `==` operator. The `==` operator applies various coercions to both sides (if they are not the same type) before testing for equality (resulting in such behavior as `"" == false` being `true`), but `Object.is()` doesn't coerce either value.

`Object.is()` is also not equivalent to the `===` operator. The only difference between `Object.is()` and `===` is in their treatment of signed zeros and `NaN` values. The `===` operator (and the `==` operator) treats the number values `-0` and `+0` as equal but treats `NaN` as not equal to each other.

<ResourceGroupTitle>Free Content</ResourceGroupTitle>
<BadgeLink colorScheme='yellow' badgeText='Read' href='https://www.scaler.com/topics/javascript/difference-between-double-equals-and-triple-equals-in-javascript/'>The Difference Between == and === in Javascript</BadgeLink>
<BadgeLink colorScheme='yellow' badgeText='Read' href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#same-value_equality_using_object.is'>Equality comparisons and sameness - MDN</BadgeLink>
```

PR #1944, opened 2022-10-01
104-accessibility 

cweave:
Added additional content to frontend accessibility.


kamranahmedse:
Thank you for the PR but let's please keep the content to minimal and focus on linking resources.

PR #1955, opened 2022-10-01
Update 105-c.md

kartikver15gr8:
Added content to learn C programming effectively for beginners!


kamranahmedse:
Thank you for the PR but let's keep the content to minmal and focus on adding resources to read more.

Issue #1838, opened 2022-09-22
Add content to JavaScript roadmap


kamranahmedse:
[JavaScript Roadmap](https://roadmap.sh/javascript) has been made interactive, but it is missing content in some of the roadmap nodes. We need to fill the [markdown files given here](https://github.com/kamranahmedse/developer-roadmap/tree/master/content/roadmaps/106-javascript/content) with a short intro to the topic mentioned in the markdown file and add some resources as [per the guidelines here](https://github.com/kamranahmedse/developer-roadmap/blob/master/contributing.md#adding-content).

**Note** — You don't have to fill all the markdown files simultaneously. Feel free to create separate pull requests with batches of changes (e.g. fill up 15 content files in each PR).

Nisarg-Chokshi:
Hello @kamranahmedse, Can I work on this ?

ZahraShahid:
Shall I work on this?

johnwinifred:
Hello, please I would love to work on this

CodeWAdi:
Hey as I am one the user  of your JAVASCRIPT roadmap.sh I would love to contribute in this issue.

hitaarthh:
@kamranahmedse Hey, I'm a JavaScript Developer, Can i Work on this?

Rohan27s:
I want to work on this issue

ojasaklechayt:
Hello!! I would like to work on the project. I have checked out the roadmap and got the chance where I can contribute. So, please assign me.

Ryrden:
Hello There, I want to work on this issue

Prashant-17-11:
Hello @kamranahmedse, I would like to work on this

Rishabh044:
Hello @kamranahmedse, Allow me to contribute on this

Arka90:
Can you please assign it to me

yashdev9274:
> Can you please assign it to me

just make the changes and send the PRs, @kamranahmedse will check whether it should get merged or not

puru-khedre:
#2132

sumitbaroniya:
Please assign me this issue.

ishratUmar18:
Can I work on this one?

sudarsan2k5:
Hello, would like to contribute.

KnightlLORD:
Hello @kamranahmedse , I would like to work on this.

c1phani1simha:
Hello @kamranahmedse, I made a pull request #2299 please check it out!

[...]

PR #1983, opened 2022-10-01
Updated Memory Life cycle in javascript

yashbhanu:
Added Information about memory life cycle in javascript
 according to issue #1838


kamranahmedse:
Can you please add some links to read more about the topic as well. Thanks!

PR #1988, opened 2022-10-02
Add roadmap+resource

lingeshwaran05:
Hey This DevOps roadmap from CommClassroom community created with many current DevOps engineers guidance and currently used by many DevOps enthusiast, It will be very helpful to Both Beginners and Persons who want to switch carrier to DevOps


kamranahmedse:
Thanks but the website that you haved linked seemed to have taken inspiration from the roadmaps we have on our website. There is nothing in their roadmaps making them worth linking.

PR #2039, opened 2022-10-02
Update 109-availability.md

ethandpark:
Adds a couple brief blurbs about the concept of availability. Added two articles that go into a few more specifics as well.


ethandpark:
Had two more for the QA roadmap I accidentally included in this pr. Deleted them and are putting them in the correct issue

PR #2043, opened 2022-10-02
Add locale to NextJS

AnTheMaker:
Hi there!
This is a small PR to add the language to NextJs, so it correctly inserts the attribute `lang="en"` to the root `<html>` element.
This should improve accessibility.

More details:
https://web.dev/html-has-lang/
https://daily-dev-tips.com/posts/nextjs-add-lang-attribute-to-html-tag/

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -37,6 +37,11 @@ let nextConfig = {
   reactStrictMode: true,
   poweredByHeader: false,
   env: loadConfig(process.env.NODE_ENV),
+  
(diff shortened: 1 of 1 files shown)


kamranahmedse:
[It isn't supported by next export](https://nextjs.org/docs/advanced-features/i18n-routing#how-does-this-work-with-static-generation) which we use for this project. Can you please add the language attribute manually to the template file? Thanks!

PR #2096, opened 2022-10-04
Added Tutorial for Solana (Blockchain Roadmap)

Jaiharishan:
Added tutorial for solana


kamranahmedse:
Thank you but there are already more than enough resources in this section

Jaiharishan:
Ok but can you consider this also.

PR #2099, opened 2022-10-04
feat: added one link in termial usage and one link in javascript conc…

Sahilnenwani:
Added the links, which I found most useful for me. Thought that these resources help other's to get clear understanding of these topics.


kamranahmedse:
Thank you but there are already more than enough resources in these sections

PR #2126, opened 2022-10-04
add error handling in go, defer, panic, recover

satboy78:


content/roadmaps/109-golang/content/100-go-basics/106-errors-panic-recover.md
```diff
@@ -1 +1,12 @@
-# Errors panic recover
\ No newline at end of file
+# Errors panic recover
+
+Discover how error handling works in Go, the list of resources below will cover:
+
+ - How an error is structured in Go
+ - How to manage errors in Go
+ - Simple strategies to have a robust and easy to debug code
```
kamranahmedse, line 7:
```suggestion
In lieu of adding exception handlers, the Go creators exploited Go’s ability to return multiple values. The most commonly used Go technique for issuing errors is to return the error as the last value in a return.

A panic typically means something went unexpectedly wrong. Mostly used to fail fast on errors that shouldn’t occur during normal operation, or that we aren’t prepared to handle gracefully.

Panic recovery in Go depends on a feature of the language called deferred functions. Go has the ability to guarantee the execution of a function at the moment its parent function returns. This happens regardless of whether the reason for the parent function’s return is a return statement, the end of the function block, or a panic.
```

content/roadmaps/109-golang/content/100-go-basics/106-errors-panic-recover.md
```diff
@@ -1 +1,12 @@
-# Errors panic recover
\ No newline at end of file
+# Errors panic recover
```
kamranahmedse, line 1:
```suggestion
# Errors/Panic/Recover
```

PR #2192, opened 2022-10-05
Added Database Sharding Subheading

SHRIYANSHGAUR:
Added a very important concept of system design related to database; Database sharding which helps to make database scalable. Hope this helps the community.


kamranahmedse:
This won't add the topics on the roadmap. But thank you for the contribution, we will get it added.

PR #2223, opened 2022-10-06
add Flamingo to Go web frameworks

Odyseya:


kamranahmedse:
Thank you for opening the issue. We don't want to convolute the roadmap by mentioning every available option and are just focusing on what is the most relevant item that you would pick today and the mentioned item doesn't conform to that. 

This has also been [mentioned in the contribution docs](https://github.com/kamranahmedse/developer-roadmap/blob/master/CONTRIBUTING.md#guidelines). 
Thank you for your suggestion though 🙏

PR #2279, opened 2022-10-08
Update 100-alchemy.md

yashdev9274:
Updated the Alchemy node of Blockchain roadmap.


kamranahmedse:
Thank you but the existing content is enough.

PR #2303, opened 2022-10-08
Added about Stack Trace and added free blog links

vermaaarti:


kamranahmedse:
Have a look at the contribution docs please; there should only be a short description (one paragraph at max) with links

PR #2304, opened 2022-10-08
Added content for Error handling using debugger including blog links …

vermaaarti:
…from GFG


kamranahmedse:
Have a look at the contribution docs please; there should only be a short description (one paragraph at max) with links

PR #2314, opened 2022-10-09
Added tutorial link for validated and attractive form with live demo

vermaaarti:


kamranahmedse:
Thank you but we already have more than enough content under forms and validation.

PR #2353, opened 2022-10-09
Issue #2112 In explicit binding call is non clickable and not highlighted on hover issue solved

### LINKED ISSUE #2112: Bug in JavaScript Roadmap
In [javascript roadmap](https://roadmap.sh/javascript), there is a small bug here. In the 'using this keyword' section, each content cannot be clicked. I hope this is fixed soon
image

parthapaul7:
close #2112 
In the javascript roadmap On **Explicit binding** hover now highlights **Call**, and **Call** is also made clickable
javascrpt.json made more consistent

Before 
image

After 
image


kamranahmedse:
Thank you for the PR but I can't merge it due to the diff size. Let me handle it. Thanks!

PR #2367, opened 2022-10-10
Clearly Indicated how JavaScript can be use inside the pure HTML files

Man07har:
I have made the required changes indicating how HTML files can use the JavaScript by just using a simple tag '<script> </script>' and can be called inside the HTML tags anytime it can by using "document.getElementById("Id_name").innerHTML = ourput_that_is_required"


kamranahmedse:
Thanks but the pre-existing detail is enough.

PR #2370, opened 2022-10-10
Further Extended the Information on the Firewall

Man07har:
Update few Information in the file and made some additions like packet filter


kamranahmedse:
Thank you but the existing detail is enough.

PR #2382, opened 2022-10-10
feat: add Kubernetes beginner tutorial

meghrathod:
Personally reviewed and recommended that this probably the best tutorial to get started in Kubernetes. It gives a basic understanding of what Kubernetes is, how it work, etc and hence a valuable addition to the roadmap.


kamranahmedse:
Thank you but we already have a lot of resources in Kubernetes.

PR #2434, opened 2022-10-12
Improved Readability

niteshxp:


kamranahmedse:
Thanks for the PR but the existing content is fine.

PR #2528, opened 2022-10-15
Update 100-terminal-multiplexers.md

AyushRaghuvanshi:
Added some properties of Multiplexing and its features for better understanding of what it is and how does it helps us.


kamranahmedse:
We are intentionally keeping the content to be a single paragraph and focusing on external resources.

PR #2600, opened 2022-10-18
added 106- javascript 118- working with API

getcoder-top:


getcoder-top:
@kamranahmedse please review

PR #2610, opened 2022-10-19
updated the python roadmap with string methods

sharmaanj200:
Added string methods in python basics.


ARYASINGHBJC:
It must be `bracket` and line 5 contains a few minor grammatical errors, including the misuse of `various different`,  `learn of`. Can you fix that!

sharmaanj200:
yes sure

PR #2619, opened 2022-10-19
Add iOS path

mchirino89:
Added iOS career path to the roaster


kamranahmedse:
The readme is only linking to the roadmaps at roadmap.sh

PR #2656, opened 2022-10-21
Updated 101-centrifugo.md

vckoushik:
Fixed the mistake from herf to href


kamranahmedse:
This has been fixed in a different PR. Thank you!

PR #2690, opened 2022-10-22
Added Search engines

SamarS1ngh:
There was no data on Search Engines, so I decided to contribute some for helping others.


kamranahmedse:
It's the development specific search engines e.g. Solr, ElasticSearch etc.

PR #2692, opened 2022-10-22
Typo Fix nodejs.json

nikhilkalburgi:
exitting --------> exiting

Please accept this with hacktoberfest tag


kamranahmedse:
Thank you for your PR but this will break the build. Changing a control name in JSON requires other changes as well. Let me handle it separately. Thanks!

PR #2712, opened 2022-10-23
Add Boot.dev's Learn HTTP course to HTTP sections

wagslane:
Adding Boot.dev's [Learn HTTP](https://boot.dev/learn/learn-http) course to the various HTTP sections (none of them have courses yet!)


kamranahmedse:
Thank you for the PR but the content seems to be behind registration and we are only adding content that is publicly available without requiring any registration.

image

wagslane:
@kamranahmedse The content is freely available without registration actually! You can click to the side of the modal and it disappears. The only difference is if you don't log in we can't keep track of progress, pass off assignments, etc. You can still access all of the content and use the code sandbox

PR #2718, opened 2022-10-23
Added gatsbyjs

SamarS1ngh:
Added more description about what Gatsby is and what does it do . Also added a link for readers to refer to.


kamranahmedse:
Thank you but the existing content is enough. We are intentionally keeping the brief introduction and provide external links to learn about the topics.

PR #2765, opened 2022-10-26
Update 105-sorting-algorithms.md

Aryasharma001:
Now we have covered all sorting algorithms in python . Added heap sort which was missing in the former links.


kamranahmedse:
The resources should fully cover the topic of the node. The link that you have shared, it only covers one piece of the topic.

PR #2768, opened 2022-10-26
content update for JS Variable Declarations

moyela:
I added content and updates to the Variable Declaration subtopic in the JavaScript roadmap


kamranahmedse:
Thank you but the existing content of the node is enough.

Issue #1885, opened 2022-09-27
Add content to DevOps roadmap


kamranahmedse:
[DevOps Roadmap](https://roadmap.sh/devops) has been made interactive, but it is missing content in some roadmap nodes. We need to fill the [markdown files given here](https://github.com/kamranahmedse/developer-roadmap/tree/master/content/roadmaps/102-devops/content) with a short introduction to the topic mentioned in the markdown file and add some resources as [per the guidelines here](https://github.com/kamranahmedse/developer-roadmap/blob/master/contributing.md#adding-content).

**Note** — You don't have to fill all the markdown files simultaneously. Feel free to create separate pull requests with batches of changes (e.g. fill up 10 content files in each PR).

hasnainmakada-99:
Hello @kamranahmedse can I work on this issue

dhruvasagar:
Created #1905 , let me know if it looks alright and I can take a look at contributing to other parts of this Roadmap.

ShobhanSrivastava:
Hey @kamranahmedse I am a beginner to open source contributing. Can I start my journey with this issue? Also can you guide me a bit on how to start contributing?

yashdev9274:
@ShobhanSrivastava go to roadmap.sh website pick any roadmap, then go to any section of that respective roadmap and click on "Edit This Page" and do changes as per your wish.

KomalMMungekar:
Hey @kamranahmedse, can I contribute to this issue? I'm a complete newbie to open source and I would like to kickstart my journey from this!

parth93QA:
Created #1972 , let me know if it looks alright.

lingeshwaran05:
Hey I have updated DevOps roadmap , itself it contains links , Docs , YT videos , courses etc please check it out

subhash-iiitm:
Created #1991 for Devops roadmap, Please take a look, let me know if it looks alright

KomalMMungekar:
Tried to update the introductory part of Ruby for DevOps roadmap by creating #2001. Kindly check it out.

yashdev9274:
Hey @kamranahmedse I created #2036  for DevOps roadmap for Jenkins section. Please have a look and let me know if any changes needed.

ethandpark:
Created [2039](https://github.com/kamranahmedse/developer-roadmap/pull/2039) for DevOps roadmap about availability in the cloud. Let me know how it looks and/or if anything needs changing!

[...]

PR #2770, opened 2022-10-27
Devops update

Karthik-59:
Added content to DevOps roadmap according to the issue: #1885


kamranahmedse:
Thanks but we are only adding English resources at the moment.

PR #2775, opened 2022-10-27
content update for setTimeout in JS

moyela:
I added more detail to the setTimeout explanation in JS


kamranahmedse:
Thank you but the existing content is enough.

PR #2788, opened 2022-10-28
Added a github repo that compares different Go Web frameworks 

pradyumnac:
Added web comparison on web frameworks node of go road map


pradyumnac:
I will submit following the content guidelines

PR #2798, opened 2022-10-28
added javascript intro resources

bytesofbec:


kamranahmedse:
Thank you but there are already more than enough resources on these nodes. Also, I don't see these courses to be extraordinarily better than the existing courses so closing the PR.

PR #2806, opened 2022-10-29
Change Prop Attribute Values

sojinsamuel:
The Anchor Text "Getting started with Angular" pointed to the Angular official docs, change the values of badgeText and colorScheme


kamranahmedse:
This has been fixed already. Thank you!

PR #2825, opened 2022-10-30
Corrected grammatically incorrect sentences, punctuations and typos for better readability. (In about.md)

Pandey-utkarsh:
Corrected grammatically incorrect sentences, punctuations and typos for better readability.


kamranahmedse:
Thank you for the PR but I don't agree with all your changes.

PR #2867, opened 2022-11-03
Update 102-native-drivers.md

techbelle:
add resources for MongoDB NodeJS driver


kamranahmedse:
We have a separate node for mongoose. This node is about using other native drivers.

PR #2924, opened 2022-11-11
docs: add Boot.dev's Learn Go course to Go resources page

wagslane:
This course is freely available without the need to sign in. The only difference is if you're not signed in you won't be able to track progress, receive achievements, etc. All the content can be read for free without sign up, and the code sandbox works without registering an account.


kamranahmedse:
Thank you for the PR, but we can't add this course.

wagslane:
No problem, could you explain why though? I'd love to update it to fit within the guidelines! @kamranahmedse

PR #2931, opened 2022-11-12
improve accessibility

Gabrielbsb21:
has improve home page accessibility.

1 - implementation of semantic tags;
2 - wai-aria implementation;


kamranahmedse:
Thank you for the PR! Although it looks good, we are working on the [next version of the roadmap.sh](https://github.com/kamranahmedse/roadmap.sh), which will be released soon. Since there are no tests at the moment, we can't see the impact of these changes, so closing this PR. But again, thank you for taking the time to contribute!

PR #2991, opened 2022-11-21
Update 100-filtering.md

Sachinkumarguri:
i found this stack blitz little tutorial very helpful for filtering the data of a table content let me know if it is a good way or i should do it some other way.
here is the link to the stack blitz https://stackblitz.com/angular/dnbermjydavk?file=app%2Ftable-overview-example.ts


kamranahmedse:
Please have a look at the contribution docs for adding content.

PR #3035, opened 2022-11-29
Adding information to the Registers and Ram node - CS Roadmap

FercueNat:


content/roadmaps/103-computer-science/content/120-how-computers-work/102-registers-and-ram.md
```diff
@@ -1,4 +1,9 @@
...
+Registers are the high-speed accessible storage elements. The processor accesses the registers within one CPU clock cycle. In fact, the processor can decode the instructions and perform operations on the register contents at the rate of more than one operation per CPU clock cycle.
+
+***Memory*** is a hardware device used to store computer programs, instructions and data. The memory that is internal to the processor is a primary memory (RAM), and the memory that is external to the processor is a secondary memory (**Hard Drive**).
+Primary memory or RAM is a volatile memory which means the data in the primary memory exist when the systems power is on, and the data vanishes as the system is switched off. The primary memory contains the data that will be required by the currently executing program in CPU. If the data required by the processor is not in primary memory, then the data is transferred from secondary storage to primary memory, and then it is fetched by the processor.
```
kamranahmedse, line 6:
```suggestion

***Registers*** are the smallest data-holding elements built into the processor itself. Registers are the memory locations that are directly accessible by the processor. The registers hold the instruction or operands currently accessed by the CPU.

Registers are the high-speed accessible storage elements. The processor accesses the registers within one CPU clock cycle. The processor can decode the instructions and perform operations on the register contents at more than one operation per CPU clock cycle.

***Memory*** is a hardware device that stores computer programs, instructions, and data. The memory that is internal to the processor is primary memory (RAM), and the memory that is external to the processor is secondary (**Hard Drive**).
Primary memory or RAM is a volatile memory, meaning the primary memory data exist when the system's power is on, and the data vanishes as the system is switched off. The primary memory contains the data required by the currently executing program in the CPU. If the data required by the processor is not in primary memory, then the data is transferred from secondary storage to primary memory, and then it is fetched by the processor.

```

PR #3048, opened 2022-12-01
design-systems glossary page unavailable

felipeluz10:
Add a snapshot of the post https://web.archive.org/web/20220620075140/https://superfriendly.com/design-systems/glossary/


felipeluz10:
The page design-systems glossary is unavailable, so I add a snapshot of the page from web.archive.org.

PR #3108, opened 2022-12-14
Fixes the `===` operator format

cr2007:
Displays it in code format so that it is appears the same as the other operators.

Currently on the website, it appears like this, looking different than the other operators displayed.
image

This PR proposes to fix this anomaly.


cr2007:
After the merge, it should appear like this
image

PR #3135, opened 2022-12-17
Update Node.js roadmap Types of Errors

YuriyVorobyov96:
Added some useful links and description to Types of Errors block


YuriyVorobyov96:
Added some useful links and description to Types of Errors block

PR #3167, opened 2022-12-22
[Build] React Fundamental Topics

sricharankrishnan:
1. Came across two other interesting topics related to props and state in react
2. Would like to suggest that we add these to the list as well as they highlight something important
3. One is from robinwieruch and the other from Dominik Dorfmeister (TkDodo)


kamranahmedse:
Thank you for the PR but we have updated the React roadmap due to which this PR can't be merged. Can you please reopen the PR with the updated roadmap? Thanks!

sricharankrishnan:
Hi kamranahmedse. Sure will have that done for you.

PR #3169, opened 2022-12-22
Update 105-javascript.md

PouriyaProgrammer:
I think you'd better to say to person who wants to use Javascript for backend that Javascript itself can't help him and he should learn additional things like Nodejs❤️


kamranahmedse:
This has been fixed. Thank you!

Issue #3179, opened 2022-12-26
Link provided for Event Loop resource in NodeJS roadmap is unavailable 


shantanupade:
First READ resource of the Event Loop (Node) in NodeJs roadmap which redirects to the NodeJs official documentation is now unavailable.

Updated URL : https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

![Screenshot_2022-12-27-01-59-16-340_com android chrome~2](https://user-images.githubusercontent.com/48827959/209582419-0c5df1bc-ae3c-43a3-baa4-cea7e13599bd.jpg)

PR #3185, opened 2022-12-27
Link updated for Event Loop resource

shantanupade:
Link updated for Event Loop resource redirecting to the official documentation in NodeJS roadmap.

Updated URL: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#what-is-the-event-loop


shantanupade:
PR for #3179

PR #3199, opened 2022-12-29
w3schools.in as an alternative to w3schools.com

therealteddy:
w3schools.in has more C concepts explained than w3schools.com

#W3SCHOOLS.IN
Screenshot 2022-12-29 at 4 32 32 PM

vs 

#W3SCHOOLS.COM 
Screenshot 2022-12-29 at 4 36 09 PM

As you can notice w3schools.in. has more information than w3schools.com


kamranahmedse:
Thank you but w3schools.com is fine.

PR #3233, opened 2023-01-04
 GraphiQL intro added 

saikumarmolleti:
Is there any thing else to add


kamranahmedse:

saikumarmolleti:
OK

PR #3240, opened 2023-01-05
Update 100-how-does-the-internet-work.md

Chris-Max-P:
I added a link to an article about the history of the internet, which nicely describes how it evolved, making clearer what is important for the functionality of the internet and how we got there.


kamranahmedse:

PR #3245, opened 2023-01-06
[Build] React Roadmap

sricharankrishnan:
1. Came across two other interesting topics related to props and state in react
2. Would like to suggest that we add these to the list as well as they highlight something important
3. One is from robinwieruch and the other from Dominik Dorfmeister (TkDodo)


sricharankrishnan:
Hey. So do have a look at these once again. I've made the PR as new as requested in our previous message. Trust this is ok for a contribution/suggestion. Have a nice day.

PR #3273, opened 2023-01-08
Fix: remove text highlight in links

TilenPogacnik:
Before:
Screenshot 2023-01-08 at 19 13 02
After:
Screenshot 2023-01-08 at 19 14 27


kamranahmedse:
Thank you!

PR #3290, opened 2023-01-12
Update theme.ts

okkesd:
that button on "https://roadmap.sh/react/" which is bottom of the page which has text of "225k GitHub Stars". I added an animation duration to it because its animation is too fast. Slower is nicer I think


kamranahmedse:
Thanks for the PR but this is intentional - we don't have any animations intentionally.

PR #3351, opened 2023-01-24
♿️ improve accessibility

Gabrielbsb21:
A contribution to improving the accessibility of the site. 

1 - implementation of semantic tags;
2 - wai-aria implementation;  [WAI-ARIA](https://www.w3.org/WAI/standards-guidelines/aria/)
3 - Using only one h1 tag per page;

thank you for your attention and patience.


kamranahmedse:
We need to work on the accessibility fixes all across the website. Closing it for now, however we do have a task for this in the backlog and will be working on it in the near future. Thanks though!

PR #3362, opened 2023-01-25
addind information about minifying CSS files

Silkiercomet:
also added an organization that do not work for profit like https://www.minifier.org/ as a tool that can be use by anyone to minify a css file


kamranahmedse:
Thanks for the PR but the "Best Practices" content is a bit different. We are working on it + the contribution guidelines.

PR #3373, opened 2023-01-27
Update 102-oracle-networks.md

mhairimcalpine:


kamranahmedse:
Thank you for the PR but we are intentionally keeping the text to be brief and focusing on providing external links for each topic.

PR #3376, opened 2023-01-27
[Feature] Dark mode

naqet:
I have implemented dark mode to the project.

I decided to use colours from the home screen for a dark mode palette.

Disclaimer: Because roadmap.sh uses SSG, there is no way to set users previously chosen theme without JS running after initial render. This causes flickering if user chooses to use theme other than the one configured on default. Because of this issue, I decided to make dark mode a default option (maybe I'm biased, but in my experience more users prefer dark mode).
Flicker white -> dark  is way worse than dark -> white. In my opinion It's not a very big issue.

This issue would be resolved if roadmap.sh could use SSR instead of raw SSG. But this is a decision to be made by maintainers, not me. I wanted to provide the best user experience with the product that we have today.

Couple of screens below: 
home_page

roadmap_example


kamranahmedse:
Thanks for the PR but we are not taking any PRs for the design changes. As for the dark mode, we don't have any immediate plans for that but we will have it in the future.

PR #3435, opened 2023-02-10
Update 103-test-oracles.md

tashdid7068:
Added few extra information on test oracle


kamranahmedse:
Thank you for the PR but the existing content is enough. We are intentionally keeping the text at the top to be brief and focusing on providing external links.

PR #3470, opened 2023-02-20
Nav/overview

naveenntj:
Complete kubernetes roadmap


kamranahmedse:
Thank you for the PR but this seems to be all written by ChatGPT. We want this content to be written some understanding of the topic and also including some external links.

PR #3475, opened 2023-02-22
Update 100-types-vs-interfaces.md

ParthGoti:
Added functionality-wise difference with code


ParthGoti:
if you like it please modify it

ParthGoti:
if you ...

kamranahmedse:
Thanks for the PR but I think the existing content is good enough.

PR #3488, opened 2023-02-24
Update 101-why-kubernetes.md

feuvpi:
Adding some text to Why Kubernetes section


kamranahmedse:
Closing since we already have this content filled up. Also this seems to be written using ChatGPT.

PR #3493, opened 2023-02-25
kubernetes-full-roadmap

naveenntj:


kamranahmedse:
Thank you for the PR but this all seems to be written by ChatGPT. We want the content in these roadmap nodes to be written by having some understanding of the topic and also including some external links.

PR #3498, opened 2023-02-26
Feat: Add resource for database indexing and SAML in backend roadmap

king-Alex-d-great:
- The database indexing resource should be included because it is one of the best and most commended videos on youtube. It breaks down the concept so that everyone (juniors and seniors) can understand it. I personally used it three years ago and I still use it with all my mentees.
- ~~Even though we have new authentication protocols, SAML is an important building block in some enterprise architectures. We should have information available for individuals looking to understand what it's all about~~. 

Update:
I noticed that the SAML part of the roadmap had been updated 2 months after my PR and so I canceled that part of the comment


genericdude1998:
Was about to put the same SAML video, one of the best resources out there to explain it

kamranahmedse:
Thank you for the PR but the link you have submitted is specifically covering SQL Server while the section is about indexing in general.

PR #3534, opened 2023-03-07
Update 105-instrumentation.md with open telemetry

mastercoder8:
Change log
---
Add open telemetry, which is widely used in the industry and is a the standard for metrics, tracing, logs


dansholds (changes requested):
Please fix the conflict and add a descriptive title i.e. "Add link to Open Telemetry docs"

PR #3543, opened 2023-03-08
removed old scrollbar for balance and better ux

YAKSHIT-22:
Old scrollbar is looking weird to this beautiful website so removed it for better ux


YAKSHIT-22:
Please have a Look

PR #3585, opened 2023-03-12
fix(react-roadmap): change react router getting started guide uri

nina-alin:
closes issue #3485


kamranahmedse:
How is this relevant to the issue you referenced?

PR #3679, opened 2023-03-19
Understanding compound index

ElanUtta:
Tring to put a better description of what indexing is and add another content about the compound index. Because the current content is a little poor.


kamranahmedse:
Thanks for the PR but the existing definition is good enough.

PR #3732, opened 2023-03-29
tooling: prettier for astro and tailwind

arikchakma:
Prettier configuration for Astro component's formatting and Tailwind CSS class sorting.


src/pages/roadmaps.astro
```diff
@@ -9,19 +9,19 @@ const skillRoadmaps = await getRoadmapsByTag('skill-roadmap');
 ---
 
 <BaseLayout
-  title='Developer Roadmaps'
+  title="Developer Roadmaps"
```
kamranahmedse, line 12: Let's keep the single quotes here

PR #3743, opened 2023-04-01
Delete 101-create-react-app.md

pedromacedol:
React team doesn't recommend CRA for new projects anymore.


kamranahmedse:
This won't delete the item from the roadmap. Thank you for the suggestion though; I will have a look.

Issue #3751, opened 2023-04-03
[Suggestion] Blockchain Developer


AdrianScarlet:
The link to "[Consensus Mechanisms in Blockchain: A Beginner’s Guide](https://crypto.com/university/consensus-mechanisms-in-blockchain)" inside the CONSENSUS MECHANISMS  chapter within Blockchain Developer Roadmap is no longer available.

AIpill:
Yes true I check it

AIpill:
I will fix it

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

PR #3752, opened 2023-04-04
Update invalid link (#3751)

AIpill:
Update the broken link to a new one from the same site


AIpill:
What's the problem? I just update the broken link.

AIpill:
Why is it failed in the build process?

PR #3773, opened 2023-04-08
Spring WebFlux item is added under the Spring core node.

amin110230:


kamranahmedse:
This will not have any impact on the roadmap. Please have a look at the contribution docs.

PR #3816, opened 2023-04-15
Remove redundancy in 101-linked-list.md

sarrietav-dev:
The deleted content is the same as 100-array.md


kamranahmedse:
This is intentional -- following linked list content builds up on that.

PR #3915, opened 2023-05-12
Add fullstack roadmap

Yash-WebDev:
added fullstack content for HTML beginners course


kamranahmedse:
Thank you but we are only focusing on the english content at the moment.

PR #3941, opened 2023-05-20
fix "Function Expression Syntax" example && add MDN links on typescript roadmap

meta-panic:


kamranahmedse:
TypeScript roadmap is the extension of JavaScript roadmap. No need to add JavaScript resources under any of the items.

meta-panic:
> TypeScript roadmap is the extension of JavaScript roadmap. No need to add JavaScript resources under any of the items.

Okay, but the example isn't accurate. Maybe I can create a new PR to fix only the example?

PR #3978, opened 2023-05-29
Add content internet

megaqbitx:
Add a link to a video that explains about the internet and I found it very educational. It is also explained by one of the fathers of the internet, Vinton Cerf. I think it's a good idea to have a video like that for people who are new to the internet.


jelnur (commented):
Checked the video. Looks good

kamranahmedse:
We have several other good resources -- I don't think this video adds any additional information to the information already present through the resources already listed.

PR #4013, opened 2023-06-05
Fix docker typo

leo-holanda:
I noticed this typo while viewing the roadmap today and decided to submit a PR to fix it.

I also took the liberty to format the docker.json file to allow more easy reviews in the future. Without this, it's not easy to search for the changes because the whole docker.json file is in just one line. I have noticed that the frontend.json file is formatted but others aren't, so if there's a reason for them please let me know so I can revert this change.

And thanks for the awesome project, it's been really helpful.

### DIFF
--- a/src/data/roadmaps/docker/docker.json
+++ b/src/data/roadmaps/docker/docker.json
@@ -272,7 +272,7 @@
           "measuredH": "25",
           "x": "703",
           "y": "1150",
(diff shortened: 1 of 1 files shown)


leo-holanda:
The typo can be found in the second commit. "Programming _Lanugages_" -> "Programming Languages". 

About the formatting, ok. I'm reverting it to the previous state.

PR #4017, opened 2023-06-07
Fix typo in docker.json

leo-holanda:
Same as #4013 but the docker.json remains unformatted as you requested.

Here's the typo in the roadmap.
Screenshot from 2023-06-07 09-59-14


amantank (commented):
LGTM

PR #4048, opened 2023-06-13
fix: Typo Lanugages -> Languages

### LINKED ISSUE #4038: [Suggestion] Docker Roadmap
Fix "Programming _Lanugages_" in web development prerequisites

arzkar:
Simple enough fix, just needed to rename the string. Fixes #4038 
I do have a question though, any reason why the json file wasn't formatted with Prettier? The whole data seems to be in just one line, are there some performance benefits of doing this that I am not seeing?


kamranahmedse:
Thank you for the PR, it is hard for me to review the change due to the nature of JSON file so I have fixed the typo myself. Regarding minified JSON, it was done to reduce the size previously but that doesn't apply anymore since we convert it to single line during the build step. I will format the JSON files. Thank you!

Issue #3312, opened 2023-01-17
Truncate/reduce repo size


Greg-Myers-SB:
Hi,
Firstly, I love the repo and site, great work!

I was cloning down your repo and I noticed it's quite a bit larger than expected.
I assume something in the past was merged and removed at a later date because the current size isn't that big.
```
 developer-roadmap$ du -h -d1
 16K	./bin
 16M	./tests
 42M	./public
 12K	./.github
983M	./.git
8.0K	./.vscode
9.5M	./src
1.0G	.
```

I used [this script](https://stackoverflow.com/a/27745221) to take a look at what was taking the space and it looks like a bunch of the same packaged nextjs blobs over and over that are no longer in the folder.
```
developer-roadmap$ ./size.bash
All sizes are in kB. The pack column is the size of the object, compressed, inside the pack file.
size  pack  SHA                                       location
3914  505   67883a514b605878a9271d156c96b8620aae0e82  _next/static/chunks/280-84bcbbc10f5bbe31.js
3914  505   e6601a1a2626a640de6b6e0c416d1a4e7858ed09  _next/static/chunks/280-f44776b3203eeefc.js
3901  503   a785adf8eaf2f3b8147fb871e322c1525d5c46de  _next/static/chunks/280-ed0412f9175fd9c3.js
3867  494   36d6bd540814579e8a5d6e5acd65ae9b8a3d75df  _next/static/chunks/280-380dee09a8fe9825.js
3828  488   80c249887e4e43466f1145ba8720bf320dcf7bdd  _next/static/chunks/280-54405b630f516751.js
3828  488   21068eefdcffcfe7ea0e0d69c3071f97548c6165  _next/static/chunks/280-5e6e7753a9bffb1c.js
3827  488   a1b453520cc5e643eb710b006649b8db566badcf  _next/static/chunks/280-5a8ed3c89d695794.js
3771  485   e16b0a7c3441071394da6526363836f47557f188  _next/static/chunks/280-82c8997798df9a61.js
3770  485   defc7756e96e8e9c941a45f8e87b5910bf5f683c  _next/static/chunks/280-b6106f7f2aab9406.js
3738  476   f404aa9696b245d748d36852f241f594dd2ff2a1  _next/static/chunks/280-c2d8454f6e138477.js
Time: 0h:02m:10s
```

Is there any desire or way to trim the repo?

This isn't a major issue, feel free to close and ignore if this isn't seen as an issue.
Regards,
Greg

[...]

PR #4130, opened 2023-06-27
Clone size fix ISSUE #3312

tabishnaqvi1311:
When cloned, the resulting download size is about 1.7gb due to the history associated with the repository. A way to get around this is to use the `depth` parameter along with the `clone` command. So the resulting command should look like

```sh
git clone --depth=1 https://github.com/kamranahmedse/developer-roadmap.git
```
This reduces the download size by a lot. The total number of objects goes from 47000 to 5000.

This has been mentioned somewhere in the readme so that contributors do not end up downloading such a large sized repository with mostly unneeded history.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -93,6 +93,12 @@ npm install
 npm run dev
 ```
 
+Note: Cloning the entire repository results in a ≈1.7GB download. If you don't need the entire history you can use the `depth` parameter to reduce clone size.
+
+```sh
+git clone --depth=1 https://github.com/kamranahmedse/developer-roadmap.git
+```
+
 ## Contribution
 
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -93,6 +93,12 @@ npm install
 npm run dev
 ```
 
+Note: Cloning the entire repository results in a ≈1.7GB download. If you don't need the entire history you can use the `depth` parameter to reduce clone size.
+
+```sh
+git clone --depth=1 https://github.com/kamranahmedse/developer-roadmap.git
+```
```
rtk-rnjn, line 100: nice

PR #4147, opened 2023-07-02
fix(typo): link break in 100-html.md

iwint:


iwint:
Last link broken in HTML section of the full stack roadmap titled "HTML and CSS Full Course - Beginner To Pro" [100-html.md](https://github.com/kamranahmedse/developer-roadmap/blob/master/src/data/roadmaps/full-stack/content/100-html.md)

PR #4243, opened 2023-07-20
Remove cache in local storage

arikchakma:
- [x] Remove local storage cache for roadmaps.


kamranahmedse:
Closing it for now, since the change in API which includes `isFavorite` seems to have fixed the inconsistent favorite marking behavior. Keeping the PR in my notes though -- if it happens again we can merge it in.

PR #4282, opened 2023-07-31
Update prompt engineering

Kimonic99:
Beam search is probably the most popular decoding algorithm for language generation tasks.
It keeps at each time step, i.e., for each new token generated, the k most probable hypotheses, according to the model used for inference, and the remaining ones are discarded.
Finally, at the end of the decoding, the hypothesis with the highest probability will be the output.

[What is Beam Search? Explaining The Beam Search Algorithm](https://www.width.ai/post/what-is-beam-search)
[Watch out for your beam search hyperparameters](https://benjaminmarie.com/watch-out-for-your-beam-search-hyperparameters/)


kamranahmedse:
Thank you for the PR but this won't add the item in the roadmap. Please have a look at the contribution docs for suggesting topics.

PR #4297, opened 2023-08-03
UI changes and alignment

vineetps:
made text changes `http` to `https` and made alignment consistent on the roadmap page


kamranahmedse:
Thank you for the PR but we are not accepting any changes to the roadmap JSON files because it is harder to test. We are working on a custom editor for roadmaps which will make it easier for people to contribute. Until then please submit an issue instead.

PR #4317, opened 2023-08-09
fix: update link of "Functional Components and Props"

smgestupa:
### Description
- Whilst following the React roadmap, the first link under "Functional Components" that points to `https://react.dev/reference/react/component`, should instead point to `https://react.dev/reference/react/components`, otherwise the user will be met with a "Not Found" page.


smgestupa:
For some reason, GitHub detected two changes where the other change has no difference, weird.

PR #4525, opened 2023-10-01
Update 100-static-typing.md

brownpanthera:
the given code was incorrect; instead of generating a compilation error, it will perform implicit conversion.


dansholds:
The code snippet has been updated since the creation of this PR and compiles with no issue.

Issue #4526, opened 2023-10-01
It's TailwindCSS not Tailwind


miguelgargallo:
### Roadmap URL

https://roadmap.sh/react

### Suggestions

It's TailwindCSS not Tailwind

avigdorsky:
Add copi laud

miguelgargallo:
Solved here: https://github.com/kamranahmedse/developer-roadmap/pull/4529

miguelgargallo:
(Close when merge)

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

PR #4529, opened 2023-10-02
Solve Tailwind > TailwindCSS

miguelgargallo:
Solve Tailwind > TailwindCSS #4526


miguelgargallo:
@avigdorsky

miguelgargallo:
the PR behind this is because your time is more valuable so I did it for you

PR #4532, opened 2023-10-02
Update 102-builtin-objects.md (duplicate url)

serverwind:
url for "Built-in Objects" is the same as "Standard built-in objects".


dansholds:
This is no longer an issue in the live roadmap.

Screenshot 2024-05-09 at 11 40 04

Issue #4361, opened 2023-08-18
AI and Data Scientist Roadmap - Deep Learning Book Sugestion


elir0d:
### Roadmap URL

https://roadmap.sh/ai-data-scientist

### Suggestions

I believe that a good alternative to the current book is https://pt.d2l.ai/index.html

It is a great book a it is fully interactive with a lot of languages available.

RajBusa:
https://github.com/kamranahmedse/developer-roadmap/pull/4556

I have considered your suggetion and have updated the back link of the ebook and is pending for review.

elir0d:
Thank you!

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

PR #4556, opened 2023-10-06
updated deep learning e book

RajBusa:
#4361

# I have personally reviewed both ebook and making a wise move by updating the ebook.

# *Why to update?*

### The https://d2l.ai/ book is widely accepted by world-class universities and also gives detailed explaination about *Transformers* which are the most hyped algorithms and in addition to that It also provides plenty of Hands-on Experience in cloud hosted jupyter notebooks (i.e Amazon Sagemaker)


kamranahmedse:
Can you please add it as a topic content instead of replacing the node link.

PR #4558, opened 2023-10-07
Fixed FIXME to use httpCall

reynadess:
Fixing the FIXME to use `httpCall` helper method instead of `fetch`


kamranahmedse:
This won't work, `httpPost` does not support form data.

Issue #4597, opened 2023-10-18
Video not available in Youtube for Python roadmap


hernanes338:
### What is the URL where the issue is happening

https://roadmap.sh/python

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Describe the Bug

In the Lambdas section, the [Python Lambda Functions](https://www.youtube.com/watch?v=kr22jigjlok) hyperlink is returning the message "This video isn't available any more" after navigating to Youtube.

The URL https://www.youtube.com/watch?v=KR22jigJLok seems to be the right one.

### Participation

- [x] I am willing to submit a pull request for this issue.

AmineAouragh:
Hey Aurelio! That's a real issue! Are you currently working on it?

PR #4598, opened 2023-10-18
Fix Python Lambda Functions hyperlink

hernanes338:
python-roadmap-lambdas URL fixed
fixes : https://github.com/kamranahmedse/developer-roadmap/issues/4597


AmineAouragh:
that's a good fix @hernanes338

Issue #4606, opened 2023-10-19
Roadmap mongoDB Data model and data type "Date"


AbdelrhmanX7:
### Roadmap URL

https://roadmap.sh/mongodb

### Suggestions

This section is incorrect based on the MongoDB documentation. You can't use $currentDate: {$type: 'date'} with insertOne. It's only available for update operations.
![Screenshot from 2023-10-19 10-49-37](https://github.com/kamranahmedse/developer-roadmap/assets/74682593/00c15d93-706a-4bab-bc29-cd0f8ef181ad)

For more details, refer to the MongoDB documentation on this topic:
https://www.mongodb.com/docs/manual/reference/operator/update/currentDate/

AmineAouragh:
Hey @AbdelrhmanX7 that's right! 

Although I don't have experience with MongoDB but after a quick google search here is what I also found 

_In MongoDB, the $currentDate operator sets the value of a field to the current date. It can be set as either a Date or a timestamp type. The default is Date. $currentDate is an update operator, and can only be used when updating documents, not when inserting them (although it can be used in upsert operations)._ - [source](https://database.guide/mongodb-currentdate/#:~:text=In%20MongoDB%2C%20the%20%24currentDate%20operator,be%20used%20in%20upsert%20operations).

I can work on that issue if there is no problem.

AbdelrhmanX7:
That's great. I will probably contribute to the Mongodb Roadmap I found some parts that need to get more clear

PR #4607, opened 2023-10-19
Fix mongodb roadmap content - change 'InsertOne' with 'Update'

AmineAouragh:
## Issue reference & explanation

This is a fix to the issue #4606 saying there is an error in the [MongoDB roadmap](https://roadmap.sh/mongodb) where an operator $currentDate (used to set the value of a field into the current date) is used in an insertOne() CRUD operation which is not correct according to this [mongodb source](https://database.guide/mongodb-currentdate/#:~:text=In%20MongoDB%2C%20the%20%24currentDate%20operator,be%20used%20in%20upsert%20operations) - the $currentDate can ONLY be used in an update() method and upsert() method.

## Details about changes:

I changed the text and the CRUD operation from 'InsertOne' to 'Update'

## Here is a preview:

2023-10-19 11_56_05-Window

### DIFF
--- a/src/data/roadmaps/ai-data-scientist/ai-data-scientist.json
+++ b/src/data/roadmaps/ai-data-scientist/ai-data-scientist.json
@@ -316,7 +316,7 @@
           "x": "279",
           "y": "583",
           "properties": {
-            "controlName": "ext_link:imp.i384100.net/baqMYv"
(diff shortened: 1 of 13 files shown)


kamranahmedse:
Closing in favor of https://github.com/kamranahmedse/developer-roadmap/pull/4630 since there are some unrelated changes in this PR.

Issue #4580, opened 2023-10-12
React Developer Roadmap: Revise React CLI tool alternative options. 


MorkWood2:
### Roadmap URL

https://roadmap.sh/react

### Suggestions

React.dev no longer recommends using the create-react-app CLI in its "Getting Started" [documentation](https://react.dev/learn/start-a-new-react-project). In fact, it was [reported](https://dev.to/ag2byte/create-react-app-is-officially-dead-h7o) to be removed from React's official documentation altogether. 

Instead, it recommends a litany of React frameworks with create-next-app at the top. 

Perhaps we can consider updating our React Roadmap to reflect those changes.

AmineAouragh:
That's right. I just checked. [Nowadays, the team recommends using production-grade React frameworks like NextJS, Remix, Gatsby for native apps.](https://www.freecodecamp.org/news/how-to-migrate-from-create-react-app-to-vite/#:~:text=The%20React%20team%20no%20longer,manage%20large%20and%20complex%20applications.)

Are you working on it? Or worked on it? 

I think I can do this if that's not a problem.

MorkWood2:
I'm not working on this one atm, go for it!

AmineAouragh:
Alright I just created a PR

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

alinkedd:
This issue should be reopened.

The last release of the `create-react-app` package was 2 years ago. The authors are unsure of its future direction. The map should mention this and definitely not refer to the abandoned and possibly broken tool with outdated dependencies as "the best way to start building a new single-page application in React."

P.S. https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741

PR #4612, opened 2023-10-20
Updated create-react-app to create-next-app in React roadmap

AmineAouragh:
According to the issue #4580 the React roadmap still has create-react-app as a React CLI tool to create new React projects. But as of now the React team no longer recommends using create-react-app but instead recommends picking one of the React-powered frameworks popular in the community like Next.js so i updated the React CLI to create-next-app.


AbdelrhmanX7:
Hey @AmineAouragh, this should be a simple task that only requires a single file change, but I noticed that around 17 files have been modified, and most of them are unrelated to the issue.

AmineAouragh:
Hey Abderahman yeah absolutely you are right. I don't understand why commits from other pull requests are being added in this pull request. There is a problem.

AbdelrhmanX7:
1- you need to sync your master fork with the developer-roadmap master branch
2- when you create a new branch do thist
    - git checkout -b your-branch-name origin/master (don't forget to stash save any uncommit files)
    - git pull origin master (to update your branch with master base)

arikchakma:
@AmineAouragh Can you please resolve the conflicting files? It could have been just one file.

alinkedd:
@AmineAouragh
Next.js is already located in `Frameworks` group.
I suggest we should either remove CRA completely or replace it with Parcel as it's mentioned in docs.

PR #4631, opened 2023-10-25
Abdelrhman: Feat auto-focused side panel

AbdelrhmanX7:
Previously, when a user opened the side panel and attempted to scroll using the arrow keys, the entire page would scroll, instead of the side panel.

Before

https://github.com/kamranahmedse/developer-roadmap/assets/74682593/25bc3605-0172-44e6-9a3a-ec3e77cff7ee

After

https://github.com/kamranahmedse/developer-roadmap/assets/74682593/f8eccb20-f6fe-4bc4-ba1a-1e346c570efb


kamranahmedse:
Thanks!

PR #4664, opened 2023-11-07
Update set.md

HenriqueMartinsBotelho:
The statement, "It is a collection of elements where each element is stored as a value without any keys," is slightly ambiguous. It's important to clarify that while `Set` does not use keys in the way that an `Object` or `Map` does, each element in a Set can be thought of as both a key and a value, which is distinct from how keys and values are defined in objects or maps.


usmanmg01:
Nokia_2.4_TA-1270_MT6750_ES2-2680-0-00WW-B02_ROM.zip
<https://drive.google.com/file/d/1LA8i2WSsQYovhdj_tiHszfQYAA17yVo_/view?usp=drivesdk>

PR #4680, opened 2023-11-11
Abdelrhman: Fix best practices articles

AbdelrhmanX7:
Before
image
After
image


AbdelrhmanX7:
i pulled the same images from gh repo but gh keep showing them as modified

PR #4722, opened 2023-11-18
Update 102-arrow-functions.md

dogukanakin:
Update 102-arrow-functions.md


dansholds:
Hi! Thanks for the PR!

When adding new content we try to add something different each time such as if the content is only text based, it's nice to add a video, or a free course etc.

This new content link is very simliar to the two already provided so I'm not going to accept the addition!

PR #4730, opened 2023-11-20
Create 103-slog.md

robertolopezlopez:
The new slog package brings structured logging to the table of the standard SDK from v1.21, thus it is important to learn.


dansholds:
Hey!

Unfortunately adding a new node to the roadmap is not as simple as creating a new file for it, although we do appreciate the effort.

Rather than continuing with this PR I will take it down as a suggestion to add a `slog` node to the roadmap, as it requires editing of the actual roadmap from our end!

robertolopezlopez:
ok :)

PR #4736, opened 2023-11-21
Correct the POSA abbreviation

bilalrahim:
The correct abbreviation for POSA is Pattern-Oriented Software Architecture not Patterns of Scalable and Adaptable Software Architecture.

Thanks


dansholds (approved):
LGTM, thanks for the catch!

PR #4741, opened 2023-11-22
QA - Fix Bug Magnet references and improve description

marcthayer:
The title of the tool mentioned should be Bug Magnet, not Bug Management. Also fixed some grammar in the description.


marcthayer:
@kamranahmedse What's your normal process for going through PRs? Should we ping you when they're ready, or do you get to them as you have time, in batches, etc.?

kamranahmedse:
Thank you for the PR but changing this would require us to write a backend data migration to update users who are tracking their progress so let's leave it as it is, since it isn't visible anywhere.

PS, sorry for the delay here -- we will be looking at the issues and PRs more regularly.

PR #4808, opened 2023-12-02
Update 101-create-react-app.md

EricBidopa:
30 days of react is also a great resource that helped me when learning react


dansholds (approved):
LGTM!

PR #4811, opened 2023-12-02
Update 102-runtime-config-options.md

iamahumannow:
Added Cgroup and Cgroup V2 descriptions

### DIFF
--- a/src/data/roadmaps/docker/content/108-running-containers/102-runtime-config-options.md
+++ b/src/data/roadmaps/docker/content/108-running-containers/102-runtime-config-options.md
@@ -44,4 +44,30 @@ Runtime configuration options allow you to customize the behavior and resources 
   docker run --hostname=my-container --dns=8.8.8.8 your-image
   ```
 
+  
+### Cgroup
+- Cgroups are a Linux kernel feature that enables the allocation and management of system resources among processes.
+- Docker leverages cgroups to provide resource isolation and limitation for containers.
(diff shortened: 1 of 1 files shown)


src/data/roadmaps/docker/content/108-running-containers/102-runtime-config-options.md
```diff
@@ -44,4 +44,30 @@ Runtime configuration options allow you to customize the behavior and resources
...
+  ```bash
+  dockerd --exec-opt native.cgroupdriver=systemd
+  ```
+- View the cgroup information with cgroup v2 enabled:
+  
+  ```bash
+  docker inspect --format '{{.HostConfig.CgroupNamespace}}' <container_id>
+  ```
+  
+Including these runtime configuration options will allow you to effectively manage your containers' resources, security, and networking needs. For a full list of available runtime configuration options, refer to Docker's [official documentation](https://docs.docker.com/engine/reference/run/).
```
dansholds, line 73: I really like the content that you've suggested here but part of me thinks just adding a link to the docker docs on Control Groups would suffice. https://docs.docker.com/config/containers/runmetrics/#control-groups

73 lines is one of our longest .md files in the project and the aim of these files is to give a brief overview and then link out to expert resources rather than posting a tutorial here.

I don't want to diminish your efforts though so I'm happy for you to also find a way to compress what you've submitted into a shorter contribution!

dansholds (changes requested):

PR #4828, opened 2023-12-06
Add resources about database transactions

HlexNC:
This pull request adds additional resources about database transactions to the project. The resources include a developer's guide from dev.to. These resources provide further information and insights into the concept of database transactions.


HlexNC:
The current N+1 query problem is only accessible after logging into medium.

PR #4840, opened 2023-12-09
Fix title for Go Roadmap

benescaglayan:


arikchakma:
This is intentional to please the lighthouse score (it complains that "Go" is not an informative button text)

PR #4874, opened 2023-12-17
Feature: highlight selected topic

EthanSKim:
Implemented by changing the color of the selected topic to rgb(255,155,0), and reverting to its original color when blank space or other topic is clicked.


kamranahmedse:
Thank you for the PR but please discuss the functionality changes before working on a feature. We don't need this.

PR #4875, opened 2023-12-17
Fix typo "Artifcats" > "Artifacts".

NeyasG:
Add small description to Artifacts and fixed a typo for folders named "Artifcats".


kamranahmedse:
Please do not change any filenames -- this will break the progress tracking.

PR #4893, opened 2023-12-22
Update AWS Best Practices in service-over-servers.md

hoalongnatsu:
Fix AWS Best Practice in section Service Over Servers point to the wrong link Chaos Monkey


arikchakma:
The link has been updated already. Thanks for the PR.

PR #4910, opened 2023-12-24
Added roadmap-list.cjs for compiling textual roadmaps

skoenig:
Helper script to generate a structured Markdown output from any roadmap. This can be useful for keeping a textual version of a roadmap in personal notes and tick off resources as learning progresses.

It has two modes - 'simple' (default) and 'full', could potentially merged with `roadmap-links.cjs` in the future.


kamranahmedse:
Thank you for the PR but there is no need for this at the moment. We might implement it in the API so that access https://roadmap.sh/frontend.txt would provide the textual version.

PR #4915, opened 2023-12-24
[C++] Weak pointer reference count clarification 

AliAshkaniNia:
I think we should mention that there is actually another counter for `weak_ptr`s. Modified the document to make it a bit more accurate.


reliek21 (commented):
👍

PR #4946, opened 2023-12-30
Added roadmap for Kotlin

Sanober494:


arikchakma:
We are now closing any issues requesting new roadmap requests to clean up the issues. 

However, this doesn't mean that the requested roadmap won't be added to the website. We are putting it in the backlog and will add it to the website depending on how it ranks in the backlog or sooner if we find someone interested in adding that roadmap. Thank you for the request though! 🙏

PR #4955, opened 2023-12-31
Fix [Issue] devops/application-monitoring/app-dynamics  show topic not found

reuteelf:
Apps Dynamics under Application Monitoring in DevOps shows topic not found

Fixed the issue


kamranahmedse:
This has been fixed in a different PR.

PR #4958, opened 2024-01-01
fix: year update in cpp.md file

bpinazmul18:
Update year in cpp.md file 2023 to 2024


dev-aly3n (commented):
Remove the year as it needs yearly change in the documents, and it also doesn't add any value

arikchakma:
The issue has been solved already.

PR #5008, opened 2024-01-09
Add depends_on to dockerfile example

TaviotBaptiste:
Hello,

Add a little information on the management of the launch order of containers in addition that can be more than convenient.

Hope it can help


dansholds (approved):
LGTM!

This content is meant to be the "basics" but I would agree that `depends_on` is an important key that can cause major problems for people if not used!

dansholds:
In the future please try to write a more descriptive PR title i.e. "Add this to that"

PR #5012, opened 2024-01-10
Update 101-as-type.md

KShehzadi:


dev-aly3n (commented):
@KShehzadi you are absolutely right.
It also could be like this:
let str = num as unknown as string;

But could you please update the file with the corrected example and include the reasoning in the pull request message?

dansholds:
This content as since been updated by another PR and is now correct.

Issue #5039, opened 2024-01-16
Please, add waitgroups to section "Going Deeper"


noo-dev:
### Roadmap URL

https://roadmap.sh/golang

### Suggestions

Please, add waitgroups to section "Going Deeper"

PR #5042, opened 2024-01-17
Add WaitGroups to Go - Going Deeper

AhmadMadi:
Adds a new - WaitGroups - node to Go roadmap, for Going Deeper. I believe it's good to have this here as it teaches about a type of concurrency in Go.

This is also done by a request [here](https://github.com/kamranahmedse/developer-roadmap/issues/5039)


AhmadMadi:
@arikchakma Hey Arik, I would appreciate it if you reviewed my PR. Thanks in advance.

arikchakma:
Please have a look at the contribution guidelines for suggesting/adding topics to existing nodes.

AhmadMadi:
@arikchakma I did have a look but I do not understand what did I miss? If it's to open an issue with my suggestion, I didn't think of it because there is already an issue which I mentioned in the PR's description (1st comment). Thanks again.

arikchakma:
The thing is, we don't allow modifying the nodes directly or adding new nodes. That's why we mentioned opening an issue with the suggestions. Thank you for your understanding.

AhmadMadi:
Oh alright, now it's clear. Thanks for the reply!

PR #5060, opened 2024-01-20
fix(signup.astro): typo unnecessary "videos" word

oscartorres-10:
fix(signup astro) typo unnecessary  videos  word


dev-aly3n (commented):
Good catch

kamranahmedse:
Thank you!

PR #5086, opened 2024-01-24
doc(PowerShell): make case and spacing consistent

bcdady:
Make text referring to PowerShell consistent with case and spacing


kamranahmedse:
This has been fixed in a different PR. Thank you!

PR #5102, opened 2024-01-28
Fix typo

little-junior:
Fix file content typo


dev-aly3n (commented):
That's not a typo. the text is quoted from W3Schools, and you can find it in the following link:
[https://www.w3schools.com/sql/sql_intro.asp](https://www.w3schools.com/sql/sql_intro.asp)

PR #5165, opened 2024-02-11
fix: navigation dropdown menu in navigation bar that is not 100% hidden

edricgalentino:
## Fix
- fix: fix navigation dropdown menu in navigation, that is not hidden if not open which it still can be click, but not anymore

### Screenshot (before fix)
- navigation open
image

- navigation closed
image

_**notice the link in the bottom left corner? it means the links is not 100% disappear which makes it still can be clicked even though the navigation dropdown is not opened.**_

### Screenshot (after fix)
- navigation open
image

- navigation closed 
image


kamranahmedse:
Thank you for the PR but this will break the animation. I have made an alternative fix.

PR #5168, opened 2024-02-11
feat: add use-freeze-scroll hook to topic detail

dev-aly3n:
The useFreezeScroll hook freezes the scroll and removes the scrollbar from the browser window when it is active, without causing any layout shifts. This is particularly useful, for example, when a modal is opened.

This pull request removes the scrollbar from the body when the topic detail is opened. It does not remove the scrollbar from the topic detail itself, so you can still scroll the content of topic while the background is locked.

I have tested this across different browsers on macOS, Linux, and Windows, and it's almost the only way to avoid layout shifts.

### DIFF
--- a/src/components/TopicDetail/TopicDetail.tsx
+++ b/src/components/TopicDetail/TopicDetail.tsx
@@ -4,6 +4,7 @@ import { useKeydown } from '../../hooks/use-keydown';
 import { useLoadTopic } from '../../hooks/use-load-topic';
 import { useOutsideClick } from '../../hooks/use-outside-click';
 import { useToggleTopic } from '../../hooks/use-toggle-topic';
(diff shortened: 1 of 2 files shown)


dev-aly3n:
fixes issue #5169 

**whitout useFreezeScroll:**

scroll2

As you can see at the right of screen there are 2 scrollbars. one for topic content and one for body.

**after using useFreezeScroll:**

scroll1

You can see the body scrollbar disappeared and background is no longer scrollable. **While it does not cause any layout shift when the body scrollbar appears and disappears.**

kamranahmedse:
Thanks for the PR but we will look into it and see if there is an alternative fix. Closing it for now since it is not that critical.

PR #5179, opened 2024-02-14
fix: disable login buttons

arikchakma:
It solves the multi-click issue, when you click any two social login buttons at a time it will throw an error. It solves the issue by disabling them when you click one of the buttons until it resolves.

### DIFF
--- a/src/pages/signup.astro
+++ b/src/pages/signup.astro
@@ -1,4 +1,5 @@
 ---
+import { AuthenticationForm } from '../components/AuthenticationFlow/AuthenticationForm';
 import Divider from '../components/AuthenticationFlow/Divider.astro';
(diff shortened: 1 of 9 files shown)


src/pages/signup.astro
```diff
@@ -29,15 +25,7 @@ import AccountLayout from '../layouts/AccountLayout.astro';
...
 
-      <div class='flex w-full flex-col items-stretch gap-2'>
-        <GitHubButton client:load />
-        <GoogleButton client:load />
-        <LinkedInButton client:load />
-      </div>
-
-      <Divider />
-
-      <EmailSignupForm client:load />
```
kamranahmedse, line 40: `AuthenticationForm` has `EmailLoginForm` and not `EmailSignupForm`

arikchakma: Ah okay let me fix it

PR #5220, opened 2024-02-20
Update 100-what-is-react-native.md

ranbir7:
Made the Introduction little catchier.


dansholds:
Thanks for the contribution but this style of copy doesn't match the rest of the site, it also doesn't add anything but rather rewrites it in a more excitable tone.

PR #5225, opened 2024-02-21
Update index.md

singhimanshu1:
Added Sql dataset image


dansholds:
This contribution doesn't add anything significant to the content unfortunately.

If we were to incorporate what you have suggested it would be via a code snippet to create the table rather than a screenshot of a table.

PR #5229, opened 2024-02-22
Update 101-edabit.md

PunitBasarkod:
"I'm a college student who updated the Edabit site with my referral link. There's no harm in that; if I earn through the referral link, it will greatly assist in covering my college fees. I understand that you may overlook this PR, but it would mean a lot to me if somehow it could provide support."


dansholds:
Hi, unfortunately, I'm not going to accept this PR because the addition of your referral link is not inline the rest of our content.

To allow one addition of a referral link would be to allow them all.

Apologies.

PR #5237, opened 2024-02-24
Update index.md

rrxmzl:
Add definition of wiki


dansholds:
Hey!

I don't believe this contribute adds value to the content given that the current link is from AWS and the youtube video covers it in depth.

PR #5262, opened 2024-02-29
Update 104-key-value-databases.md

mimi0007:
Added Redis in Action Book by Josiah L. Carlson which is a great resource to learn Redis.


dansholds:
This content is Redis specific and would belong in the Redis section not the overarching KV DB section.

PR #5269, opened 2024-02-29
Update 103-technical-skills.md

alexlatam:
Php Language added to list of programming languages. Software architect roadmap


dansholds:
Hey!

Thanks for the contribution but I'm not going to merge it purely out of principle that we could keep adding to it i.e. Go, Kotlin, Bash etc

PR #5321, opened 2024-03-10
Update 101-function.md

akarsanth:


src/data/roadmaps/javascript/content/101-javascript-variables/103-scopes/101-function.md
```diff
@@ -1,6 +1,6 @@
 # Function Scope
 
-When a variable is declared inside a function, it is only accessible within that function and cannot be used outside that function.
+When a variable is declared inside a function, it is only accessible within that function and cannot be used outside that function. This scope applies to variables declared using the let, const, and var keywords.
```
dansholds, line 3: This is not true as let and const are block scoped.

dansholds:
Due to the inaccuracy of the addition I'm going to decline this PR, given that if you were to correct my comment, the remaining contribution would be too small to be considered impactful.

PR #5331, opened 2024-03-12
Update 100-what-are-containers.md

Gudisa1:
I have Added a scenario for what's Container Content


dansholds:
Hey!

Unfortunately this `Scenario` type content is not in line with the style of content that we have in the roadmaps, especially as the entire Docker roadmap serves as this scenario for containers!

PR #5332, opened 2024-03-12
Add Bun as a task runner

bobralks:
add bun for running package.json scripts


dansholds (approved):
Bun is great! and has a fantastic OS community!

PR #5347, opened 2024-03-14
Update index.md

ddonggi:
Assign a name to unnamed content


dansholds:
While I agree that we should have _some_ content in this section, the majority of the links that you've suggested already exist in the sections for those specific areas i.e. the Nginx link is present in the Load Balancer section.

PR #5394, opened 2024-03-22
Update 100-database-design-basics.md

A9B9:
Adding a book resource for databases


dansholds:
Hey!

Thanks for the PR but we don't accept contributions of paid content such as purchasable books or in this case, one you can access with a 10 day free trial.

PR #5409, opened 2024-03-25
Update 100-render-equation.md

Hendo10X:


kamranahmedse:
Please use markdown format for content

PR #5413, opened 2024-03-27
Update 103-n-plus-one-problem.md to add non-paywalled link

jbdotjs:
## Change:

Replaced [**Medium**](https://medium.com/doctolib/understanding-and-fixing-n-1-query-30623109fe89) member-only article with [**freeCodeCamp**](https://www.freecodecamp.org/news/n-plus-one-query-problem/) alternative having similar quality and content.

## Problem:

The original article has paywalled access and hence should be removed.


dansholds (approved):
LGTM Thanks!

dansholds:
I have just noticed that the paywalled link has since been replaced already! Thanks for the PR but I will close the PR due to that.

PR #5414, opened 2024-03-27
Update 100-osi-model.md

VinayPrabhakar-gamer:
OCI text corrected to OSI under Application Layer section


VinayPrabhakar-gamer:
I think the OSI is wrongly written to OCI under Application layer section.

arikchakma (approved):

Issue #5109, opened 2024-01-30
Outdated URL


Altaner:
### What is the URL where the issue is happening

https://github.com/kamranahmedse/developer-roadmap/blob/master/src/data/roadmaps/backend/content/109-apis/104-hateoas.md

### What browsers are you seeing the problem on?

Microsoft Edge

### Describe the Bug

On this page, the YouTube URL is outdated. The URL is not working.

### Participation

- [x] I am willing to submit a pull request for this issue.

dansholds:
No outdated Youtube URL found.

PR #5438, opened 2024-03-31
Added HATEOAS resources

belikedeep:
fixes : #5109

As i checked there was no Outdated URL as such. 

Also added an interesting article on HATEOAS


dansholds (changes requested):
Hey Deepak, This is a create addition to a node that was previously lacking any content but please squash your commits and I will approve.

dansholds (approved):

PR #5460, opened 2024-04-03
Replace article link

JDeepD:
The previous URL redirects to a whole different website which has nothing related to the topic.


dansholds (approved):
Valid contribution, the old link went to their homepage and not to an article!

PR #5474, opened 2024-04-07
Update 100-bj-frogg-behavior-model.md

ccxdcyl:


dansholds:
Hey thanks for the PR!

The style that you've implemented doesn't match our style of section content, we typically just bold and bullet point sections (as you can see in the other sections of the roadmap!).

The title change is also unnecessary as the content is on the UX Design roadmap, so no need to declare that again.

PR #5486, opened 2024-04-10
Added detailed explanation and example of static typing in C++

N-E-W-T-O-N:
In the current version, it was mentioned that assigning a double type variable's value to int type variable leads to a compile time error, which is false. This will lead to internal `type conversion` .
The flow of type conversion is as follows 
`char`->`int`->`float`->`double` .
Moving right to left lead to data loss while moving left to right have no data loss .


N-E-W-T-O-N:
Screenshot from 2024-04-11 00-19-23

PR #5543, opened 2024-04-23
feat: add user open graph

arikchakma:


src/pages/u/[username].astro
```diff
@@ -26,10 +22,16 @@ let errorMessage = '';
 if (error || !userDetails) {
   errorMessage = error?.message || 'User not found';
 }
+
+const ogImage = `${import.meta.env.PUBLIC_API_URL}/v1-profile-open-graph/${username}`;
```
Comment (kamranahmedse) [inline src/pages/u/[username].astro:26]: Let's add a server rendered route in roadmap.sh for the og image. This way it will get cached in cloudfront as well.
API urls won't get cached.

kamranahmedse (changes requested):

src/pages/u/[username].astro
Comment (arikchakma) [reply src/pages/u/[username].astro:26]: But we can cache the response using `Cache-Control` or no?

PR #5571, opened 2024-04-28
chore: replace with og endpoint

arikchakma:


src/pages/og/[slug].ts
```diff
@@ -4,13 +4,22 @@ import { getDefaultOpenGraphImageBuffer } from '../../lib/open-graph';
...
 export const GET: APIRoute<any, Params> = async (context) => {
   const { slug } = context.params;
 
-  if (!slug.startsWith('user-')) {
+  if (
+    !slug.startsWith('user-') &&
+    !slug.startsWith('roadmap-') &&
+    !slug.startsWith('best-practice') &&
+    !slug.startsWith('guide-')
+  ) {
```
Comment (kamranahmedse) [inline src/pages/og/[slug].ts:22]: I was thinking about this, let's split the routes for each i.e.

```
https://roadmap.sh/og/user/[username]
https://roadmap.sh/og/roadmap/[slug]
https://roadmap.sh/og/best-practice/[slug]
https://roadmap.sh/og/guide/[slug]
```
Because each of these will have a separate caching duration `/og/user/*` being minimal and others being aggressively cached.

kamranahmedse (changes requested):

src/pages/og/[slug].ts
Comment (arikchakma) [reply src/pages/og/[slug].ts:22]: Okay

PR #5582, opened 2024-05-01
Fix type in copy

shto:
fix incorrect sentence


dansholds (approved):
Good catch!

PR #5611, opened 2024-05-07
Add Boot.dev's "Learn python" course to relevant roadmaps

hbeckmann:
This adds the ["Learn Python" course](https://boot.dev/learn/learn-python) from Boot.dev to relevant roadmaps.

You can freely access all content without registering on Boot.dev. By clicking the "browse the lessons" link, users get access to 186 python lessons.


kamranahmedse:
Thank you for the PR. There was already a PR from Lane to add the courses from boot.dev. We can't add these unfortunately since boot.dev is a paid product (even though this particular resource isn't) and this popup that keeps showing up every now and then.

image

hbeckmann:
> Thank you for the PR. There was already a PR from Lane to add the courses from boot.dev. We can't add these unfortunately since boot.dev is a paid product (even though this particular resource isn't) and this popup that keeps showing up every now and then.
> 
> image

Thanks for the reply @kamranahmedse!  I work alongside Lane at Boot.dev and thought this would be a valuable resource for people starting their journey in Python. We realize that the free and open nature of your project makes it extremely valuable for everyone, I reviewed the contribution guidelines and thought that we were in line with the requirements - is there any more clarification we could get on the matter? I notice that Codecademy is on the roadmap and that they force you to create an account for their course.

We do a lot of sponsorships and work with other free resources (ie. FreeCodeCamp) and would love to find any way we can make changes to include our content on this project. Thanks again!

Issue #3741, opened 2023-04-01
[Suggestion] Vue Developer


skirtles-code:
Some changes you might consider:

* `nextTick` is not part of the Composition API. I'd consider not including `nextTick()` at all, as it seems no more important than several other topics that aren't mentioned.
* Vue CLI has been replaced. `create-vue` is now used for scaffolding, which uses Vite. Vitest is a popular choice for unit testing.
* I would remove Vuepress and Gridsome and add VitePress instead.
* Vue Formulate was for Vue 2. It's now FormKit.
* VueUse is a really important part of the Vue ecosystem.
* The API Calls section seems a bit out of place. Several of the libraries listed there are orthogonal to using Vue. Some are very rarely used in practice.
* If props are seen as fundamental then I think custom events and slots are too. Arguably component `v-model` deserves a separate mention too. I would probably have scoped slots listed separately, as an advanced topic. If the Fundamentals section feels a bit crowded, I think Attribute Inheritance could be moved down to Advanced.

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

PR #5621, opened 2024-05-08
Added warning to Vue CLI regarding its maintenance mode

artshade:
Dear Developers,

Thank you very much for the awesome resource and project!

The `create-vue` recommendation was added [2 years](<https://github.com/vuejs/vue-cli/commit/7c0134e653a0f58315ac9b06e940a5c5928b8b0b>) ago by Evan You, and it feels like it must be stated in the feature description here, too.

Best and kind regards :sparkles:

### DIFF
--- a/src/data/roadmaps/vue/content/100-fundamental-topics/100-vue-cli.md
+++ b/src/data/roadmaps/vue/content/100-fundamental-topics/100-vue-cli.md
@@ -1,5 +1,7 @@
 # Vue CLI
 
+> ⚠️ **\[Warning]** Vue CLI is now in maintenance mode. For new projects, please use [create-vue](<https://github.com/vuejs/create-vue>) to scaffold [Vite](<https://vitejs.dev/>)-based projects.
+
 Vue CLI is a full system for rapid Vue.js development, providing:
 
 - Interactive project scaffolding via `@vue/cli`.


dansholds:
Hey!

Thanks for creating the PR!

You raise an interesting point but I'm conscious that the link within that content which goes to https://cli.vuejs.org/guide/ does show this warning at the top.

Do you feel like adding a warning here (bare in mind that we don't have deprecation warnings or emojis anywhere else) is really necessary?

dansholds:
It seems like the solution here lays overall updates to the Vue roadmap rather than adding this depreciation warning doesn't it!

I'll close this PR and raise these points internally!

Thanks for raising this, we really appreciate it!

skirtles-code:
@dansholds For what it's worth, we frequently get questions about learning Vue on the Vue Discord server, and occasionally the link to this roadmap gets suggested. But invariably we have to discourage people from using it, because the information is too out of date and would lead beginners in the wrong direction.

If it helps, I made some suggestions a while back about some of the things that might be worth changing: #3741.

dansholds:
> @dansholds For what it's worth, we frequently get questions about learning Vue on the Vue Discord server, and occasionally the link to this roadmap gets suggested. But invariably we have to discourage people from using it, because the information is too out of date and would lead beginners in the wrong direction.
> 
> If it helps, I made some suggestions a while back about some of the things that might be worth changing: #3741.

Ah that's awesome, thanks! 

Feedback like this is invaluable so I'll make sure it's fed back to the team and hopefully we can see some of your suggestions implemented 🤙

I'll keep you updated via the linked suggestion!

PR #5623, opened 2024-05-09
fix: double scroll bar when searching

comehere127:
image

Fix a minor bug on ui. 
In the homepage, when user click search, search modal appeared but in the right conner has double scroll bar.


dansholds:
Hey @comehere127 !

What browser are you experiencing this bug on? I can't replicate on four different browsers 😅

comehere127:
> Hey @comehere127 !
> 
> What browser are you experiencing this bug on? I can't replicate on four different browsers 😅

i got issue on safari, arc, firefox

arikchakma:
This is a legit issue, I am having a look at the PR. Thanks @comehere127.

arikchakma (approved):

PR #5624, opened 2024-05-09
Fix typo in input section of API Security Best Practice Roadmap

dansholds:
**Before**:

User proper HTTP methods for the operation

**After**:

Use proper HTTP methods for the operation


dansholds:
Screenshot 2024-05-09 at 09 23 51

PR #5632, opened 2024-05-09
Fixed link display of 101-why-rust.md

BeardedOwl1357:
Unnecessary tabs makes markdown render them as "code" instead of links

(Thank you so much for this website btw)

# Currently Looks like this on website

image

# Before

image

# After

image


dansholds (approved):
Good catch, and thanks for the kind words!

PR #5685, opened 2024-05-16
Update 102-metro-bundler.md

igvishalpandey:
Added a Article on Metro Bundler and its uses in React Native


kamranahmedse:
We are intentionally keeping the topic details short and focusing on providing external resources for learning.

PR #5687, opened 2024-05-17
Minor update in readme.md file

salmantec:
## Description:

- Updated readme.md file with `cd developer-roadmap` command to move into that folder before running further npm command.

### DIFF
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -45,7 +45,10 @@ const videos = await getAllVideos();
       featuredItems={skillRoadmaps
         .filter((roadmapItem) => !roadmapItem.frontmatter.isHidden)
         .map((roadmapItem) => ({
-          text: roadmapItem.frontmatter.briefTitle,
+          text:
+            roadmapItem.frontmatter.briefTitle === 'Go'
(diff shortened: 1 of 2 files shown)


src/pages/index.astro
```diff
@@ -45,10 +45,7 @@ const videos = await getAllVideos();
       featuredItems={skillRoadmaps
         .filter((roadmapItem) => !roadmapItem.frontmatter.isHidden)
         .map((roadmapItem) => ({
-          text:
-            roadmapItem.frontmatter.briefTitle === 'Go'
-              ? 'Go Roadmap'
-              : roadmapItem.frontmatter.briefTitle,
+          text: roadmapItem.frontmatter.briefTitle,
```
kamranahmedse, line 48: This is intentional to make lighthouse happy, otherwise it considers that as an action button and throws warning that "Go" is not a good name for a button. Please revert.

kamranahmedse (changes requested):

src/pages/index.astro
salmantec: Reverted this changes. Please check it now.

PR #5710, opened 2024-05-21
Update api-design.json

09liweis:
fix Reate Limiting to Rate Limiting typo


dansholds:
Hey @09liweis! Thanks for the catch!

We don't accept PRs that make huge changes to the JSON as it's very hard to validate, I'll fix this myself 🤝

PR #5714, opened 2024-05-22
Added additional links to the data structures and algorithms roadmap

anishomsy:
I took the links from the computer science roadmaps and added them to the data structures and algorithms roadmaps. I think it would be better to have links as additional resources instead of just texts.


dansholds:
Hey!

While we appreciate the time this would have taken, these changes are unnecessary.

You have added the same links to multiple files as well as taken the number of links in some files to excess.

Please refer to our contribution guide for a maximum number of links and the statement "Adding everything available out there is not the goal!". There is an significant number of links that we _could_ add to the roadmaps but that would only overwhelm our users rather than help them.

PR #5716, opened 2024-05-22
Fixed typo in Docker roadmap

Skyth3r:
Found a duplicate 'defines' for the Distribution Specification (distribution-spec) section so I've just removed that now 👍🏽


dansholds (approved):
Excellent spot! 10 points to Griffindor! 🦁

PR #5719, opened 2024-05-23
modify the sample code

zongsforce:
the original sample code's complexity is O(n) instead of O(n!)


dansholds:
Correct me if I'm wrong @zongsforce but your example is factorial, not exponential, we have provided our (beginner) users with a simple example of a function with exponential time complexity.

zongsforce:
> Correct me if I'm wrong @zongsforce but your example is factorial, not exponential, we have provided our (beginner) users with a simple example of a function with exponential time complexity.

Sorry, I made a mistake in the code. Thanks for pointing it out. I have submitted a new PR.

PR #5725, opened 2024-05-24
The tip float can be hidden by the user.

comehere127:
https://github.com/kamranahmedse/developer-roadmap/assets/11963675/3bfcf59f-8a3e-4cda-95f0-f7f35d9f7509


kamranahmedse:
Thank you for taking the time to open the PR. 
Looks really nice but I have to close it unfortunately, we have an alternative implementation planned for this.

PR #5735, opened 2024-05-26
Function Pointer's desciption

Aterolite24:
Initialization of Function Pointer clarified


kamranahmedse:
We are intentionally keeping this content minimal and linking to external resources.

Aterolite24:
Got it, thanks.

PR #5745, opened 2024-05-28
feat: add useful links for learning generics in golang

pixel365:


dansholds (approved):
Awesome resources thanks!

Issue #5741, opened 2024-05-27
Mention the RustRover IDE in "IDEs and Rust Toolchains"


fellalli:
### Roadmap URL

https://roadmap.sh/rust

### Suggestions

https://github.com/kamranahmedse/developer-roadmap/blob/ba630173b8d87dc4b8bbfee99a607ca06489bac0/src/data/roadmaps/rust/content/100-introduction/104-ides-and-rust.md?plain=1#L3

`IntelliJ-based IDEs also support Rust through their plugin.`

It might be a good idea to mention [RustRover](https://www.jetbrains.com/rust/), which was released a few days ago, as it is a full IDE rather than just a plugin.

kamranahmedse:
Thank you, feel free to open a PR.

PR #5748, opened 2024-05-28
Mentioned RustRover in "IDEs and Rust Toolchains" (#5741)

fellalli:
The Rust plugin by JetBrains is deprecated. Instead, there is [RustRover](https://www.jetbrains.com/rust/), a dedicated IDE for Rust development.


dansholds (approved):
LGTM thanks for keeping this up to date :)

PR #5749, opened 2024-05-28
Update 100-create-vue.md. Add a resource

nikitadmitr:
Add an article about how "npm create vue@latest" works under the hood. Beginners often get confused about this command


kamranahmedse:
Thank you for the PR. The article looks interesting but I don't think we should be recommending this as a resource for someone who is just getting started with Vue. When user is on this node, we want them to create a Vue project and move on. Rest of these topics and in-depth coverage can be introduced later on the roadmap. We have some plans to improve on this future and will add such content then. Thanks though!

PR #5755, opened 2024-05-29
add copy and links to architecture topics

dansholds:
Add copy and links to the **Core OS**, **Core Services** & **Cocoa Touch** topics.


kamranahmedse (changes requested):
Let's just add a brief introduction (1 or 2 paragraphs at max) for each and add external links.

dansholds:
> Let's just add a brief introduction (1 or 2 paragraphs at max) for each and add external links.

Done 👍 🦖

PR #5757, opened 2024-05-29
feat: add Leptos framework to the web development section of the Rust roadmap

pixel365:
Leptos is a very popular and actively developing powerful framework.

I suggest recommending it for study.

https://github.com/leptos-rs/leptos


dansholds:
Hey @pixel365 thanks for yet another solid contribution!

We're currently paused on making changes to the roadmap like this (new nodes etc) as we are moving everything over to our Editor!

(Doesn't mean we won't add this though)

PR #5769, opened 2024-05-30
Add `Required` Utility in Typescript

ItsmeBlackOps:


dansholds:
Please see my comment on your previous PR, you cannot add a new topic by simply adding a new file.

This file is also 2/3x larger than the other utility type content so when making contributions please bear in mind the styling of the other elements in the roadmap to ensure it matches.

PR #5784, opened 2024-06-01
fix: correct spelling from "Synatx" to "Syntax"

dansenpir:
Changed the spelling from "Synatx" to "Syntax" to improve readability and maintain consistency across the topic.


dansholds (approved):
Yikes! Good catch @dansenpir!

PR #5787, opened 2024-06-01
Fix: Standardize using "docker container ls" command when referencing listing containers

### LINKED ISSUE #5783: Standardize using docker ps or docker container ls -a in listing containers
### Roadmap URL

https://roadmap.sh/docker

### Suggestions

In the Running Containers section, the Listing Containers description mentions using `docker ps` to list containers whereas the example uses `docker container ls -a`. Can we change the description or example to be consistent and use the same command in both places. Maybe just have another note to mention the existence of the other option to show they both do the same thing.

apun97:
Replaced instances of `docker ps` with `docker container ls`

Resolves #5783


dansholds (approved):
LGTM

PR #5822, opened 2024-06-07
fix: remove broken import

pixel365:
"import { isMobile } from '../../../editor/utils/is-mobile.ts';" - it is invalid path

This method is located in '../../lib/is-mobile.ts', but it is not used in this component either


dansholds:
Can you address the removal of the golang content tags please, probably best to rebase your branch and resubmit.

PR #5823, opened 2024-06-07
Update 109-Diff bw Pressable and TouchableOpacity .md

sravim589:
I have added to know what are the diff bw Pressable and ToucableOapcity, when we should used as per its need.


dansholds:
This section is specifically about `Pressable`, mentioning other methods could be confusing and off-topic.

sravim589:
Okay

PR #5833, opened 2024-06-07
Update resource link for OOP section

JesusG16:
One of the links for the introduction to object oriented programming no longer leads to the site it used to but to an unrelated one, I found a link I think would be suitable for this.


dansholds (approved):
LGTM thanks

PR #5888, opened 2024-06-18
Update 102-update.md

karim1safan:
update the salary to 45000$ not 5000$


fellalli (commented):
No, 5000 is correct.

The old value is 40000. If we add 5000 to it, then it's 45000.

In the SQL example, the value is set to 45000, which increases the old value by 5000.

```sql
UPDATE Employees
SET Salary = 45000
WHERE EmployeeID = 3;
```

dansholds:
Correct as per @fellalli comment.

PR #5915, opened 2024-06-24
Add nose2 into python testing

shengchuang-wong:
Add nose2 into python testing


kamranahmedse:
This will not have any impact on the roadmap. Please have a look at the contribution docs for recommending topics.

PR #5970, opened 2024-06-29
feat(roadmap/angular): add signals content

Kordrad:


dansholds:
Hi @Kordrad simply creating a new markdown file will not create an interactive topic on the roadmap, if you believe that Signals is an integral part of Angular basics then we can look to add it in the future once we have migrated the roadmap to the new editor.

Kordrad:
> Hi @Kordrad simply creating a new markdown file will not create an interactive topic on the roadmap, if you believe that Signals is an integral part of Angular basics then we can look to add it in the future once we have migrated the roadmap to the new editor.

souds good. I wasn't sure how the new element should be created. Thanks for providing the information.

PR #5979, opened 2024-06-30
fix: [roadmaps/full-stack] Change "full stack" to "full-stack" to be more precise

duyphuongn:
This PR contains the following changes:

- All "full stack" instances are replaced with "full-stack" to be more precise as "full-stack" is a hyphenated noun compound, and "Full-stack Developer" title should be preferred.

- Improvements: "step by step" to "step-by-step", "notes app" to "note app", "project based" to "project-based"...


dansholds (changes requested):
While agree with this in certain situations lets stick to just changing content files and not the files that control the SEO of the roadmap or the JSON

kamranahmedse:
Thank you for the PR but we will take a look at this ourselves as we need to assess the impact. Closing it for now.

PR #5981, opened 2024-06-30
Addition: [roadmaps/DataAnalyst] Add article for Kurtosis

AmiraliToori:
Add a link (article) that describes the topic of Kurtosis, and I think it might be useful to understanding the concept better.


dansholds:
LGTM

PR #5985, opened 2024-06-30
Fixing broken rxjs link

kaellandrade:
- Fixing broken rxjs link roadmap Angular.

image


dansholds:
LGTM, this is an old roadmap with old link formating but this should still be actioned 👍

Issue #5987, opened 2024-06-30
Article to understand 3 Trees and Keys in Flutter


MJ04-20:
### Feature Description

In the Flutter Developer Roadmap we see no resource for the topic - 3 Trees.
I'd like to submit an article to be put in here. 
Link - [Flutter Internals](https://medium.com/@subroto.2003/a-guide-to-flutter-internals-ce8d64d01c50)
I request you to check it. The article received a very warm response from both beginners and experienced flutter devs, I believe this can be helpful for anyone starting out with flutter.
@kamranahmedse

dansholds:
Article looks great, please submit this as a PR.

PR #5995, opened 2024-07-01
feat: Added Article for flutter internals

MJ04-20:
Added the Flutter Article for Flutter Internals. 
As discussed with @dansholds 
Under Issue #5987


MJ04-20:
@dansholds Sir,
May I know when it will be published on the website?

dansholds:
@MJ04-20 within 24 hours. We only do one build per day.

PR #6035, opened 2024-07-04
Update iboutlets@tuUuLInq0p-nhehe2AqPg.md

mrgsdev:


src/data/roadmaps/ios/content/iboutlets@tuUuLInq0p-nhehe2AqPg.md
```diff
@@ -1 +1,3 @@
-# IBOutlets
\ No newline at end of file
+# IBOutlets
+An outlet is a property of an object that references another object. The reference is archived through Interface Builder. The connections between the containing object and its outlets are reestablished every time the containing object is unarchived from its nib file. The containing object holds an outlet declared as a property with the type qualifier of IBOutlet and a weak option.
+- [@official@Outlets](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Outlets/Outlets.html)
```
dansholds, line 3:
```suggestion
# IBOutlets

An outlet is a property of an object that references another object. The reference is archived through Interface Builder. The connections between the containing object and its outlets are reestablished every time the containing object is unarchived from its nib file. The containing object holds an outlet declared as a property with the type qualifier of IBOutlet and a weak option.

Learn more from the following resources:

- [@official@Outlets](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Outlets/Outlets.html)
```

PR #6057, opened 2024-07-04
Update 100-what-is-rust.md

bahomaid23:
Adding a new learning resource to what is Rust.

The new added resource is a video by fireship on YouTube, which explains Rust in 100 seconds.


dansholds:
@bahomaid23 This video is already present in the "Why use Rust" topic.

PR #6064, opened 2024-07-05
Update content-negotiation@TX_hg7EobNJhmWKsMCaT1.md

BLACK00HOLE:
Added content in API design for "Content of Negotiation"


dansholds:
Sorry, but this doesn't add anything significant to the content. A better contribution would be links to relevant resources.

PR #6103, opened 2024-07-08
fix: checkbox right click

### LINKED ISSUE #6090: Right-clicking the checkbox marks the checkbox as done but doesn't mark the topic as done
### What is the URL where the issue is happening

https://roadmap.sh/best-practices/frontend-performance

### What browsers are you seeing the problem on?

Chrome

### Describe the Bug

Right clicking the checkboxes on **_best-practices_** roadmaps only marks the checkbox as done and does not strike off the topic name.

It also doesn't update the progress analytics on the top of the page, it only updates the progress analytics box on the bottom of the page

Ref Image: 
image

### Output from browser console (if any)

_No response_

### Participation

- [ ] I am willing to submit a pull request for this issue.

arikchakma:
Resolves #6090

### DIFF
--- a/src/components/FrameRenderer/renderer.ts
+++ b/src/components/FrameRenderer/renderer.ts
@@ -190,7 +190,9 @@ export class Renderer {
     e.preventDefault();
 
     const isCurrentStatusDone = targetGroup.classList.contains('done');
-    const normalizedGroupId = groupId.replace(/^\d+-/, '');
+    const normalizedGroupId = groupId
+      .replace(/^\d+-/, '')
+      .replace('check:', '');
 
     if (normalizedGroupId.startsWith('ext_link:')) {
       return;
--- a/src/components/FrameRenderer/renderer.ts
+++ b/src/components/FrameRenderer/renderer.ts
@@ -152,6 +152,10 @@ export class Renderer {
       return;
     }
 
+    if (/^check:/.test(topicId)) {
+      topicId = topicId.replace('check:', '');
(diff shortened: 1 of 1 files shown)


src/components/FrameRenderer/renderer.ts
```diff
@@ -196,6 +196,19 @@ export class Renderer {
...
+          detail: {
+            topicId: groupId.replace('check:', ''),
+            resourceType: this.resourceType,
+            resourceId: this.resourceId,
+          },
+        }),
+      );
+      return;
+    }
+
```
kamranahmedse, line 211: This can simply be done inside the `updateTopicStatus` -- just remove `check:` from the start of the group ID before making the API call.

kamranahmedse (changes requested):

src/components/FrameRenderer/renderer.ts
arikchakma: Lol, that's what I did at first, then I saw the `click` code. After that, I modified it😆

src/components/FrameRenderer/renderer.ts
```diff
@@ -152,6 +152,10 @@ export class Renderer {
       return;
     }
 
+    if (/^check:/.test(topicId)) {
```
kamranahmedse, line 155: It's fine for now but something for future: try to avoid regex for simple cases like this. It could be `topicId.startsWith('check:')`

PR #6113, opened 2024-07-09
FIx links for sql vs nosql in system-design and sql roadmaps

AllainPL:
Wanted to also make sure is there some guidance on similar content? Should those 2 files be copies of the same content? Or is there some other approach


dansholds:
@AllainPL they're on totally seperate roadmaps so this is fine, if the links are good it's ok to have them appear across roadmaps 👍

PR #6121, opened 2024-07-09
Update 110-linked-list.md

hiahir357:
A link was added to the LinkedList step of the Rust roadmap


dansholds:
@hiahir357 you've added the same link as the link that is already in this topic.

PR #6160, opened 2024-07-16
fixed React SSR option name to Remix

willopesweb:
the label in SSR -> React was being presented as "react-router" but the content was about Remix. I changed the label name to "Remix"


kamranahmedse:
[Thank you for the PR but this is intentional.](https://remix.run/blog/merging-remix-and-react-router)

PR #6184, opened 2024-07-19
added Dash to Python frameworks

omnicoder9:
Created 103-dash.md.


dansholds:
@omnicoder9 This isn't how new interactive nodes are added to the roadmap, creating the file doesn't make the topic node in the JSON.

We may add Dash but the idea isn't to add every framework or module but rather enough to give people what they need to go forward and learn using other resources.

I would suggest adding a link to dash in the "framework" topic for the time being.

PR #6199, opened 2024-07-20
Fix persistant search result

### LINKED ISSUE #6193: Duplicate Search Results
### What is the URL where the issue is happening

https://roadmap.sh/

### What browsers are you seeing the problem on?

Firefox, Chrome, Microsoft Edge, Other

### Describe the Bug

Certain roadmaps, including `Code Review`, `Backend`, `Node.js`, and `React`, appear multiple times when the same term is searched repeatedly.

https://github.com/user-attachments/assets/cc77257f-8920-454f-8720-3df24ef665cf

### Output from browser console (if any)

_No response_

### Participation

- [ ] I am willing to submit a pull request for this issue.

mr-loop-1:
fixes #6193 

the problem was that multiple groups had same .md filename resulting in same id and hence key in fragment. Hence, I made it a composite key.

### DIFF
--- a/src/components/CommandMenu/CommandMenu.tsx
+++ b/src/components/CommandMenu/CommandMenu.tsx
@@ -243,7 +243,7 @@ export function CommandMenu() {
                 const groupChanged = prevPage && prevPage.group !== page.group;
 
                 return (
-                  <Fragment key={page.id}>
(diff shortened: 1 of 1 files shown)


kamranahmedse:
Thank you!

PR #6223, opened 2024-07-24
Update dns.md 

andreluis-oliveira:
Fix the YouTube video link error and update the video…


andreluis-oliveira:
Update understand-cia-triad.md

1 - Add Professor Messer’s YouTube video.
2 - Remove the Varonis article link.

Note: The Varonis article does not contain any additional relevant content, so the text of this project alone is sufficient.

PR #6226, opened 2024-07-25
Update 100-text.md

Avicii786:
I update the code for displaying the Alert in react native.


Avicii786:
Update the code for Alert in react native when click on text.

dansholds:
the existing code snippet is perfectly valid.

PR #6278, opened 2024-07-30
fix: interpolation typo in filename

jdegand:


kamranahmedse:
Filename should match the node title in the roadmap. Can you please fix the typo in the roadmap also?

jdegand:
I fixed the typo in `angular.json`.  I think that's it.

PR #6285, opened 2024-07-30
Ensure when TopicDetail is open document.body stops scrolling

Ne-Ne:
As per the video below, I noticed that when scrolling around the pull out UI the body scroll position was lost.

A bit annoying, seems like a simple fix.

Before:
https://github.com/user-attachments/assets/a349541c-1f54-415d-bdaa-8eb808749b80

After:
https://github.com/user-attachments/assets/ae49c77a-936b-4bf8-8118-956298050127


kamranahmedse:
Thank you!

Ne-Ne:
> Thank you!

No problem, @dansholds said PR's were open 🚀

PR #6291, opened 2024-07-31
fix: remove duplicate change-detection topic angular roadmap

jdegand:
Change Detection as a subtopic of `pipes` makes no sense.  I deleted the one `changeDetection` with no content and changed the `angular.json` to point to other change detection file.


dansholds:
@jdegand appreciated but this isn't the best approach to handling this.

Also changeDetection as the metadata item and Change Detection with Pipes are two different topics 👍

PR #6306, opened 2024-07-31
Fix compilre typo in Typescript roadmap

craigbwagner:
Apologies for the previous pull request. Although I read the contribution page, I'm a new developer and misunderstood what I was looking at (particularly the JSON file at first read). Is this the correct way to fix the issue?


kamranahmedse:
Yes, thank you! 🙏

PR #6318, opened 2024-08-01
[Fix] Increased username length to 20

### LINKED ISSUE #6248: Allow usernames with more than 10 characters
### Feature Description

I wish I could use both my name and my surname as my username, but the field only allows 10 characters.

prateeksingh19:
Fixes #6248. User wasn't able to give username more than 10 characters.


src/components/UpdateProfile/ProfileUsername.tsx
```diff
@@ -119,7 +119,7 @@ export function ProfileUsername(props: ProfileUsernameProps) {
               // only allow letters, numbers
               const keyCode = e.key;
               const validKey =
-                USERNAME_REGEX.test(keyCode) && username.length <= 10;
+                USERNAME_REGEX.test(keyCode) && username.length <= 20;
```
Kordrad, line 122: You can improve the code so that no  _magic numbers_  are used

src/components/UpdateProfile/ProfileUsername.tsx
```diff
@@ -131,7 +131,7 @@ export function ProfileUsername(props: ProfileUsernameProps) {
             }}
             onInput={(e) => {
               const value = (e.target as HTMLInputElement).value?.trim();
-              const isValid = USERNAME_REGEX.test(value) && value.length <= 10;
+              const isValid = USERNAME_REGEX.test(value) && value.length <= 20;
```
Kordrad, line 134: Looks like duplicate code. Also, it could be improved.

Kordrad (commented):
I added some suggestions 😄

src/components/UpdateProfile/ProfileUsername.tsx
prateeksingh19: yeah, i'll look into it tmrw. im having this issue where i wasnt able to login or signup as a user to check these limitations once ran the project locally idk why, could you help me with it pls??

Kordrad: I meant that above you have such code: 

` const validKey = USERNAME_REGEX.test(keyCode) && username.length <= 20;`

and here the same statement. 

I have two solutions, how to improve code:

### I. Improve USERNAME_REGEX

You don't need always check `USERNAME_REGEX.test(value) && value.length <= 20;` 
In 60th line you can make regex like this:

```diff
- const USERNAME_REGEX = /^[a-zA-Z0-9]*$/;
+ const USERNAME_REGEX = /^[a-zA-Z0-9]{0,20}$/;
```

then only use `.test(value)`
```ts
const isValid = USERNAME_REGEX.test(value);
```

here is an example:
https://regex101.com/r/XVbgdc/1


### II. Create function to check

```diff
const USERNAME_REGEX = /^[a-zA-Z0-9]*$/;
+const isUserNameValid= (value: string) =>  USERNAME_REGEX.test(value) && value.length <= 20;
```

and you don't need to create constants like this:

```ts
const isValid = USERNAME_REGEX.test(value);
```

but this will be enough 
```ts
if (!isUserNameValid(value)) {
 return;
}
```
--- 
As for the question about logging in locally, I will not be able to help you, I do not have the knowledge on this subject yet, I think @dansholds will be able to help you.

prateeksingh19: ok, i'll look into implementing it. thanks for the help

prateeksingh19:
i did the changes you told me about and also reduced some redundant codes.

package-lock.json
```diff
@@ -64,11 +64,11 @@
...
         "markdown-it": "^14.1.0",
-        "openai": "^4.52.2",
-        "prettier": "^3.3.2",
-        "prettier-plugin-astro": "^0.14.0",
+        "openai": "^4.52.7",
+        "prettier": "^3.3.3",
+        "prettier-plugin-astro": "^0.14.1",
         "prettier-plugin-tailwindcss": "^0.6.5",
-        "tsx": "^4.16.0"
+        "tsx": "^4.16.2"
```
Kordrad, line 71: Changes in package-lock.json probably should be done in different PR.

Kordrad (commented):
Looks good to me, but I think package-lock.json should be not updated in this PR. Next time you can just run `npm ci`

package-lock.json
prateeksingh19: sry, i didnt know about it, will keep in mind from now on, thank you.
should i fix it? or is it fine?

Kordrad: idk 😄 @kamranahmedse before merge will provide information about this

kamranahmedse: Thanks for reviewing @Kordrad 🙏 
@prateeksingh19 yes please, let's remove the package-lock.json.

prateeksingh19: i have fixed it, @Kordrad kindly review it, if it needs any more changes. Thank You.

Kordrad (commented):
Looks good 👌

Issue #6331, opened 2024-08-02
Testing mindset -> QA mindset


403-html:
### Roadmap URL

https://roadmap.sh/qa

### Suggestions

As we know testing is just part of quality assurance topic. And the roadmaps itself is for quality assurance engineer, not manual tester.  
 

![image](https://github.com/user-attachments/assets/e836ac91-d015-48ab-84de-e32a11f495dc)

Overall, the description seems to focus heavily on a "reactive" mindset, but it would be more beneficial to adopt a holistic and proactive approach to quality.

While resources like ISTQB offer useful guidance, they often expect an unrealistic level of perfect alignment and adherence. Real-world QA should consider the human element, understanding that mistakes are bound to happen, and instead of aiming to "100% prevent them", the focus should be on designing processes to minimize their impact. It's important to view defects as opportunities for growth rather than blaming developers.

The idea that QAs are the sole core of finding defects is misleading, as testing can also be done by developers or project managers. The "Tester Mindset" section needs a significant overhaul. As a Quality Assurance Engineer, your role is to advocate for quality throughout the software development lifecycle, creating an environment where everyone can do their best work, especially during critical moments like a hotfix. If you notice a teammate deeply focused, offering to bring them a coffee can be a simple yet supportive gesture to help them maintain their concentration.

From my point of view "Testing mindset" should be redone. I can try and write something up.

kamranahmedse:
Feel free to submit a PR. Thank you!

PR #6344, opened 2024-08-02
docs: update testers mindset docs to QA mindset docs

403-html:
This PR updates and expands a bit docs about mindset of Quality Assurance Engineer, as this position is much more broad, than merely testing stuff (old docs are just about manual testing, I'd even say that it's about only exploratory testing).

# Changes

*   updated mapping
*   updated docs from reactive description, to proactive and holistic approach (this is more what should be done in organizations)
*   added more soft skills
*   added automation as one of key points which is nice to grow in (in some point you should)
*   updated key points (left some/rephrased them)
*   removed obsolete ISTQB syllabus (kept the fresh one, but I still think it's more as a "what it can be" but in reality we should weight it and not go 100% with it)

Related issue (more about why I think this change was needed): #6331

note: as "non maintainer" I can't use `npm run generate-renderer`? I get that I don't have permissions to some repo.


kamranahmedse:
Thank you!

PR #6364, opened 2024-08-04
docs(angular): add @if content

jdegand:
This could be expanded further later on.  But this is a decent start.


src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
-# @if
\ No newline at end of file
+# @if
+
+The @if block conditionally displays its content when its condition expression is truthy.  Content is added and removed from the DOM based on the evaluation of conditional expressions in the @if and @else blocks.
+
+```
```
Kordrad, line 5: example shlould be marked as HTML:

```suggestion
```HTML
```

src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
...
+const count = signal(0);
+
+@if(count){
```
Kordrad, line 8: if you use signals, you should to use `()`. This example will always be true. Don't provide invalid examples into description.

src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
-# @if
\ No newline at end of file
+# @if
+
+The @if block conditionally displays its content when its condition expression is truthy.  Content is added and removed from the DOM based on the evaluation of conditional expressions in the @if and @else blocks.
```
Kordrad, line 3:
```suggestion
The @if block conditionally displays its content when its condition expression is truthy. Content is added and removed from the DOM based on the evaluation of conditional expressions in the @if and @else blocks.
```

src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
...
+
+```
+const count = signal(0);
```
Kordrad, line 6: You can't use `const` in HTML

src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
...
+
+@if(count){
+    return true;
+}
+```
+
+Remember to read a signal inside a conditional because a signal will always be evaluated as truthy since signals cannot be null.
```
Kordrad, line 13: I think, we don't need information about signal in `@if` roadmap.

src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,28 @@
...
+
+```
+const count = signal(0);
+
+@if(count() === 1){
+    return false;
+}
+```
```
Kordrad, line 21: Also, I think this will be a better example:

```html
<div class="user-container">
  @if (x > y) {
    <b>{{x}}</b> is greater than <b>{{y}}</b>
  } @else if (y > x) {
    <b>{{x}}</b> is less than <b>{{y}}</b>
  } @else {
    <b>{{x}</b>} is equal to <b>{{y}}</b>
  }
</div>
```

because the HTML tags suggest for user that `@if` statement should be used inside `html` file.

jdegand:
Well, I'm too tired and bungled this.

PR #6371, opened 2024-08-05
docs(angular): add @if content

jdegand:
@Kordrad I bungled the other PR, but I incorporated your feedback into this and maybe this is better and potentially setup to be expanded further later.

### DIFF
--- a/src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
+++ b/src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
@@ -1,1 +1,25 @@
 # @if
+
+The @if block conditionally displays its content when its condition expression is truthy. Content is added and removed from the DOM based on the evaluation of conditional expressions in the @if and @else blocks.
+
+'''html
+<div>
+  @if (x > y) {
+    <p>{{x}} is greater than {{y}}</p>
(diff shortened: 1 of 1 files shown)


src/data/roadmaps/angular/content/if@U1Zy2T-2ki9pDkXn9hn-I.md
```diff
@@ -1 +1,25 @@
...
+    <p>{{x}} is equal to {{y}}</p>
+  }
+</div>
+'''
+
+Some things to be mindful of when using @if:
+
+- If you forget to read a signal, the conditional will always be evaluated as truthy since signals cannot be null.
+- Inside an @if conditional, using optional chaining on an object's property can cause TypeScript issues.
+
```
kamranahmedse, line 21: Let's just stick to a single paragraph and external resources
```suggestion
'''html
<div>
  @if (x > y) {
    <p>{{x}} is greater than {{y}}</p>
  } @else if (y > x) {
    <p>{{x}} is less than {{y}}</p>
  } @else {
    <p>{{x}} is equal to {{y}}</p>
  }
</div>
'''

Some things to be mindful of when using @if:

- If you forget to read a signal, the conditional will always be evaluated as truthy since signals cannot be null.
- Inside an @if conditional, using optional chaining on an object's property can cause TypeScript issues.
```

PR #6410, opened 2024-08-07
Update 101-volume-mounts.md

Hebe-330:
Add a link to help beginners understanding the subtle differences between `-v` and `--mount` flag.

### DIFF
--- a/src/data/roadmaps/docker/content/104-data-persistence/101-volume-mounts.md
+++ b/src/data/roadmaps/docker/content/104-data-persistence/101-volume-mounts.md
@@ -33,6 +33,7 @@ docker run -d --mount source=my-volume,destination=/data your-image
 ```
 
 In both examples above, `my-volume` is the name of the volume we created earlier, and `/data` is the path inside the container where the volume will be mounted.
+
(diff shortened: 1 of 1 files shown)


src/data/roadmaps/docker/content/104-data-persistence/101-volume-mounts.md
```diff
@@ -33,6 +33,8 @@ docker run -d --mount source=my-volume,destination=/data your-image
 ```
 
 In both examples above, `my-volume` is the name of the volume we created earlier, and `/data` is the path inside the container where the volume will be mounted.
+> For an in-depth exploration of the `-v` and `--mount` flags, consult Docker's official guide on [Choose the -v or --mount flag](https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag).
```
kamranahmedse, line 36:
```suggestion

> For an in-depth exploration of the `-v` and `--mount` flags, consult Docker's official guide on [Choose the -v or --mount flag](https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag).
```

PR #6459, opened 2024-08-09
docs(linux): wrangle content and add content links to user management

siamonas:
Wrangled content and added contents links to the three (3) sub-topics of **User Management**


dansholds (changes requested):
Lets focus on a few things here:

1. The written copy needs to be a single, concise paragraph that gives a high level overview of the topic
2. Remember to put "Learn more from the following resources" before the link section
3. We have alot of geeks4geeks content on the platform, lets diverse those links a little

Thanks!

siamonas:
Thank you for your feedback, @dansholds.

PR #6515, opened 2024-08-12
Fix sets@soZFqivM3YBuljeX6PoaX.md

bigtendons8:
Ugh... okay, looks there was ANOTHER error in my last pull request... i think this fixed it??

Really sorry about this... just wanted to add some resources :P
I probably should have created an error for this but I panicked and did this in a rush as you can probably tell from the commit history


kamranahmedse:
No worries, thank you for the fix 🙏

PR #6535, opened 2024-08-13
update half of the content

dansholds:
committing what I have so far, I will loop back to this roadmap after Git roadmaps.


src/data/roadmaps/sql/content/ceiling@BAqJQvcguhIhzyII5LRH6.md
```diff
@@ -1,24 +1,7 @@
...
-FROM Products;
-```
-
-In this example, if the original price was $10.25, the `RoundedUpPrice` will be $11. This is because the `CEILING` function rounds up the 'Price' value to the nearest whole number.
-
-It's essential to remember that `CEILING` always rounds up. So even if the Price is $10.01, the RoundedUpPrice according to `CEILING` would still be $11. If you want to round to the nearest whole number, you might want to use the `ROUND` function instead. 
-
-Another important note is that the return type of `CEILING` will be of the same type as the provided numeric expression. For instance, if you supply a numeric expression of type decimal, the return type will also be of type decimal.
\ No newline at end of file
+- [@article@CEILING - SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/functions/ceiling-transact-sql?view=sql-server-ver16)
```
kamranahmedse, line 7: T-SQL is the propriety and is extension to SQL. Let's try to find alternative resources instead of the resources from microsoft t-sql docs.

src/data/roadmaps/sql/content/abs@6vYFb_D1N_-tLWZftL-Az.md
```diff
@@ -1,38 +1,8 @@
...
-| OrderID | Product | Absolute Quantity |
-|---------|---------|-------------------|
-| 1       | Apple   | 5                 |
-| 2       | Banana  | 10                |
-| 3       | Cherry  | 15                |
-
-As you can see, the negative values in the 'Quantity' column have been converted to positive values by the `ABS` function.
\ No newline at end of file
+- [@article@ABS - SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/functions/abs-transact-sql?view=sql-server-ver16)
+- [@article@ABS - PostgreSQL](https://www.w3schools.com/sql/sql_ref_sql.php)
```
dansholds, line 8:
```suggestion
- [@article@ABS](https://www.w3schools.com/sql/func_sqlserver_abs.asp)
```

PR #6537, opened 2024-08-13
Update 102-types-and-type-assertions.md

git-amw:
I've add more explanation and relevant links


kamranahmedse:
We intentionally keep it to 3 to 4 good links.

git-amw:
Should i remove the extra links ? or we are keeping it as is without any updates ?

PR #6567, opened 2024-08-14
Update template-literal-types@xP25nGw42VqdfZ_9pDMXd.md

frandsoh:
This makes the backtick render correctly. I noticed in different browsers that the double backtick would render as “ instead of two single backticks next to each other.


kamranahmedse:
Thank you but this is intentional

frandsoh:
@kamranahmedse But you do realize it renders wrong on the website, right? 
image

PR #6568, opened 2024-08-14
Update decorators@TRx3wvWnRdJJf5eL-GBG4.md

frandsoh:
Add note about using correct compiler options.


kamranahmedse:
We intentionally keep this content to a single paragraph.

PR #6576, opened 2024-08-14
updated Cross-Join Content in the SQL Roadmap

AniketGupta30:


kamranahmedse:
Please have a look at the contribution docs. Content should [follow this format](https://raw.githubusercontent.com/kamranahmedse/developer-roadmap/master/src/data/roadmaps/react/content/react-native%40NvXAq1vN2wpncdW-yTL4c.md), i.e., one paragraph explaining the topic and a list of resources (all in markdown format).

AniketGupta30:
I think the content that I posted was according to the self-join section, so it can be valid.

PR #6578, opened 2024-08-14
Updating index.md with new C++ article

KananHasanov747:
This website provides a decent amount of knowledge of C++ (including SDL), like learncpp.com. Don't get confused by AI images, it's just author had fun playing with them.


kamranahmedse:
Thanks for the PR but we don't accept any links that require user registration.

KananHasanov747:
> Thanks for the PR but we don't accept any links that require user registration.
Actually it doesn't. It's a matter of preference. You can still read the whole material for free.

PR #6599, opened 2024-08-16
docs(java): remove nb! in threads

jdegand:
I don't know what this `NB!` was supposed to be for. If it is supposed to alter the markdown, it isn't working.


kamranahmedse:
NB = [Nota Bene](https://en.wikipedia.org/wiki/Nota_bene) i.e. "Important note"

But I agree, we can remove it.

PR #6632, opened 2024-08-17
Updated select-star.md

pyoneerC:
This PR adds [Best Practices For SQL Query Optimizations](https://www.geeksforgeeks.org/best-practices-for-sql-query-optimizations/). It helps people understand better how to practically apply the advice shown in the page.


kamranahmedse:
The content in this node should be about "SELECT *" only. The linked resource is a general set of best practices.

PR #6658, opened 2024-08-18
6th -> 5th most starred

pyoneerC:
This PR changes all '6th' to '5th'

image

PR #6674, opened 2024-08-18
Script that updates the repo ranking

pyoneerC:
We could use an automation to get the ranking of the repository.

_Originally posted by @siamonas in https://github.com/kamranahmedse/developer-roadmap/issues/6658#issuecomment-2295224476_

Draft because I have no idea how to changes variables in a `.astro` with the script, help appreciated.


.github/workflows/ranking.yml
```diff
@@ -0,0 +1,20 @@
+name: Check and update GitHub Ranking in page
```
kamranahmedse, line 1: I would rather recommend doing this during the build, [just like we get the discord member count](https://github.com/kamranahmedse/developer-roadmap/blob/master/src/lib/discord.ts).

kamranahmedse (changes requested):

.github/workflows/ranking.yml
siamonas: Sounds good!!

@pyoneerC If you'd like to tackle this on your own, that's great! Otherwise, I'd be happy to help or take it on for you.

pyoneerC:
Nice @siamonas

PR #6680, opened 2024-08-18
Updated graphics-api@Kx7O7RLp7aPGtOvK8e314.md

pyoneerC:
Adds a video comparing directx vs vulkan. (vulkan better)


dansholds:
While it's a fun video to see a graphics comparison, it's not really adding any learning resources to the topic.

PR #6688, opened 2024-08-19
Update GitHub repository rank display automatically

siamonas:
Based on discussion under #6674 

All other files, except one, seem to produce the expected result! I am not able to test the `generate-og-images.mjs`.

Feedback is always appriciated!


kamranahmedse:
Thank you, there are some changes needed but I will push it myself 🙏

PR #6707, opened 2024-08-20
error fixed

mohit15-web:


mohit15-web:
in navbar after clicking on menu toggle is not working i have fixed that isseu

kamranahmedse:
Please revert your changes in package-lock.json

PR #6767, opened 2024-08-23
Update JS string content - added backticks ( `` )

iliAsghar:
added backtick ( `` )


src/data/roadmaps/javascript/content/string@1RcwBHU3jzx0YxxUGZic4.md
```diff
@@ -1,6 +1,6 @@
 # String
 
-String is a primitive type that holds a sequence of characters. String in Javascript is written within a pair of single quotation marks `''` or double quotation marks `""`. Both quotes can be used to contain a string but only if the starting quote is the same as the end quote.
+String is a primitive type that holds a sequence of characters. String in Javascript is written within a pair of single quotation marks `''`, double quotation marks `""`, or backticks ` `` ` (template literals). all types of quotes can be used to contain a string but only if the starting quote is the same as the end quote.
```
kamranahmedse, line 3:
```suggestion
String is a primitive type that holds a sequence of characters. String in Javascript is written within a pair of single quotation marks `''`, double quotation marks `""`, or backticks ` `` ` (template literals). All types of quotes can be used to contain a string but only if the starting quote is the same as the end quote.
```

PR #6806, opened 2024-08-25
Moved content from Typecasting topic to Exception topic

RaulGalvez288:
Roadmap edited: Python

Topics edited: Typecasting, Exceptions

Before the pull request:
image
image

Description: 
I moved content that had to do with exceptions from the Typecasting topic to the Exceptions topic.


kamranahmedse:
This file is autogenerated. Any bugs should be solved in the relevant content files inside the roadmap directory.

PR #6819, opened 2024-08-27
Improvements to ssl-vs-tls.md

andreluis-oliveira:
- rename ssl-vs-tls.md@6ILPXeUDDmmYRiA_gNTSr.md to http-vs-https@6ILPXeUDDmmYRiA_gNTSr.md
- 
Note: The reason for renaming the file is because of the title of the section (secure vs. insecure), so it makes more sense to use http vs. https than two secure protocols like ssl vs. tls.

- New content


dansholds:
You don't need to edit the migration file, and please clean up the commits in this PR.

PR #6846, opened 2024-08-27
remove link loc from 102-storage-volume.md

stmoody:
the link to Elastic Block Store documentation has a hard coded language that causes it to open to the Portuguese version

removing this will default to the users preferred language


dansholds:
Good spot

PR #6890, opened 2024-08-30
Update frontend.json

ujjwaljha1:


andreluis-oliveira:
irrelevant, the Internet Fundamentals video and course is better than the mozzila article.

Please refer to the [contribution guidelines](https://github.com/kamranahmedse/developer-roadmap/blob/master/contributing.md)

- Rewriting content in a way that doesn't add any value.

PR #6897, opened 2024-08-31
add real examples about segmentations

sergiomarotco:
Hi, want to add some real examples of network segmentations.


dansholds (changes requested):
While these additions are good they don't follow our contribution guidelines, i.e we dont have images in the topics etc

Please refer to the contribution guidelines to see the format that we use, a simple fix would be to preserve your links and preface them with "Learn more from the following resources:"

sergiomarotco:
@dansholds hi! Now is good?

dansholds (approved):
Close, but not quite.

I've remediated but check out what I changed and do that in future :)

Issue #6888, opened 2024-08-30
Dart article link not working - Variables


aryanraw:
### Roadmap URL

https://roadmap.sh/flutter

### Suggestions

![image](https://github.com/user-attachments/assets/7d977de9-357f-4046-a1e0-ca496848e8a5)

I suggest changing the link to something else, such as [this link](https://learnxinyminutes.com/docs/dart/).

PR #6931, opened 2024-09-01
Fix: Comment out `howtoflutter.dev` (site down) and update tag from `…

SOUMITRO-SAHA:
Fix:
Commented out the howtoflutter.dev link (site down) and updated the tag from @article@Variables to @official@Variables #6888.

### DIFF
--- a/src/data/roadmaps/flutter/content/100-dart-basics/101-variables.md
+++ b/src/data/roadmaps/flutter/content/100-dart-basics/101-variables.md
@@ -9,5 +9,5 @@ Variables in Flutter can store values of different data types, such as numbers, 
 
 Visit the following resources to learn more:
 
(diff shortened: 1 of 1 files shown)


kenshanta:
Better remove the commented code.

Not sure if leaving comments in the code is allowed or not so I'll open a question/suggestion regarding this topic

SOUMITRO-SAHA:
I left the comment because, in the future, when the site goes live or if they want to add something else in that area, it could be helpful. I can remove it if any contributor or maintainer requests it. Since there hasn’t been any discussion on this matter, I decided to leave it for now, but I’m happy to make any changes if needed.

Thank you, @kenshanta , for bringing up this matter for further discussion.

kenshanta:
I've opened a question issue #6934 regarding this topic.

dansholds:
An outright deletion is the method we want to use 🤙

PR #6970, opened 2024-09-02
Fixed a typo on the word "visualisation" - (visualization).

datosconyam:
notice that file name is also written wrong and affect the way it's displayed on the screen.

image


kamranahmedse:
We will update it on the roadmap. Thank you!

PR #6980, opened 2024-09-03
Update usememo@w3bNp7OkehI1gjx8NzlC8.md

saeid1994:
Memorizes" vs. "Memoizes": The term "memorizes" can imply simply remembering something, while "memoizes" is a specific term in computer science meaning to cache the result of a function. Using "memoizes" is more accurate in this context.

### DIFF
--- a/src/data/roadmaps/react/content/usememo@w3bNp7OkehI1gjx8NzlC8.md
+++ b/src/data/roadmaps/react/content/usememo@w3bNp7OkehI1gjx8NzlC8.md
@@ -1,6 +1,6 @@
 # useMemo
 
-`useMemo` is a React hook that memorizes the result of a function. It is used to optimize performance by caching the result of a function and returning the cached result when the inputs to the function have not changed.
+`useMemo` is a React hook that memoizes the result of a function. It is used to optimize performance by caching the result of a function and returning the cached result when the inputs to the function have not changed.
 
 Learn more from the following resources:
 


src/data/roadmaps/react/content/usememo@w3bNp7OkehI1gjx8NzlC8.md
```diff
@@ -1,6 +1,6 @@
 # useMemo
 
-`useMemo` is a React hook that memorizes the result of a function. It is used to optimize performance by caching the result of a function and returning the cached result when the inputs to the function have not changed.
+`useMemo` is a React hook that memoizes the result of a function. It is used to optimize performance by caching the result of a function and returning the cached result when the inputs to the function have not changed.
```
arikchakma, line 3:
```suggestion
`useMemo` is a React hook that caches the result of a function. It is used to optimize performance by caching the result of a function and returning the cached result when the inputs to the function have not changed.
```

arikchakma: It would be clearer if we used `cache` instead of `memoize`.

saeid1994: @arikchakma memoizes is a specific term in computer science meaning to cache and as you see in this sentence in those paragraph it describes more : "It is used to optimize performance by caching..."

arikchakma: Makes sense

arikchakma (approved):

Issue #6966, opened 2024-09-02
Docker Roadmap - Tags are not displayed properly


split-sec:
### Detailed Description

![image](https://github.com/user-attachments/assets/c66796e2-59a0-48bb-af2b-d1b153abb2b8)

There are these issues in 3 parts in Docker Roadmap - Installation / Setup, Bind mounts, Union Filesystems.

I believe that the issue is due to the fullstop added at the end of each line.

fellalli:
Same issue for *roadmap.sh/cpp -> Setting up your Environment*. I suppose that resources are currently only detected correctly 
 when there is no additional text in the same line.

PR #6982, opened 2024-09-03
Fix/6877 - Wrong markdown in Docker Roadmap - Union Filesystems' content

split-sec:
After fixes local looks like this

image


split-sec:
I think this also fixes #6966

PR #6984, opened 2024-09-04
update scrum@PKqwKvoffm0unwcFwpojk.md

novinbukannopin:
added a description and added some reference links about scrum


dansholds:
This has already been solved by another PR, but thanks for taking the time.

PR #7024, opened 2024-09-06
Update 101-auth-logs.md

Apapora:
Added info on how to find logs for systems using systemd


kamranahmedse:

PR #7044, opened 2024-09-08
Update specular@odfZWKtPbb-lC35oeTCNV.md

ENIOmega:


src/data/roadmaps/game-developer/content/specular@odfZWKtPbb-lC35oeTCNV.md
```diff
@@ -0,0 +1,5 @@
+# Specular
+
+`Specular` reflections are mirror-like reflections. In these cases, the rays of light are reflected, more than they are absorbed. The angle of incidence is equal to the angle of reflection, that is to say that the angle at which the light enters the medium and then bounces off, the angle of the beam that bounced off would be the same.
+
+-[Specular reflection](https://www.youtube.com/watch?v=2cFvJkc4pQk)
```
dansholds, line 5:
```suggestion
`Specular` reflections are mirror-like reflections. In these cases, the rays of light are reflected, more than they are absorbed. The angle of incidence is equal to the angle of reflection, that is to say that the angle at which the light enters the medium and then bounces off, the angle of the beam that bounced off would be the same.

Learn more from the following resources:

-[@video@Specular reflection](https://www.youtube.com/watch?v=2cFvJkc4pQk)

```

PR #7056, opened 2024-09-09
6953 angular remove xsrf protection topic

dansholds:
removed the XSRF protection topic as it was a dupe of another topic.

### DIFF
--- a/src/data/roadmaps/angular/angular.json
+++ b/src/data/roadmaps/angular/angular.json
@@ -257,8 +257,7 @@
       },
       "dragging": false,
       "selectable": true,
-      "focusable": true,
-      "resizing": false
+      "focusable": true
     },
     {
       "id": "nyJvnCmX-pC35SbkRPXko",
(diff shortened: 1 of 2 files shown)


src/data/roadmaps/angular/angular.json
```diff
@@ -210,10 +210,10 @@
         "y": 2260.93042937083
       },
       "width": 277,
-      "height": 249,
+      "height": 196,
       "style": {
         "width": 277,
-        "height": 249
+        "height": 196
       },
```
arikchakma, line 217: Let's not modify the JSON directly otherwise the `editor` state and code state will not be the same. We can keep notes and remove them in the next iteration.

cc @kamranahmedse

dansholds: These changes where done via the desktop tool that we have, sometimes it randomly changes the height of things and I usually catch it and revert it. I'll close this PR for now then and we can remove it at a later date.

PR #7058, opened 2024-09-09
turned blank button into DevOps Roadmap button

dansholds:
In the "Learn to Automate" section, we had a blank blue button just called "button", the assumption is that it was going to link to the DevOps roadmap.

### DIFF
--- a/src/data/roadmaps/postgresql-dba/postgresql-dba.json
+++ b/src/data/roadmaps/postgresql-dba/postgresql-dba.json
@@ -578,7 +578,7 @@
     },
     {
       "id": "2zqZkyVgigifcRS1H7F_b",
-      "type": "DevOps Roadmap",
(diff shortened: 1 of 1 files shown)


src/data/roadmaps/postgresql-dba/postgresql-dba.json
```diff
@@ -578,15 +578,15 @@
     },
     {
       "id": "2zqZkyVgigifcRS1H7F_b",
-      "type": "button",
+      "type": "DevOps Roadmap",
```
kamranahmedse, line 581: This doesn't seem right. `type` can only be button

arikchakma, line 581:
```suggestion
      "type": "button",
```

I think it will button not DevOps Roadmap

PR #7142, opened 2024-09-17
devops networking fix ping typo

darthwalsh:
You can't ping a URL -- that will just confuse a learner

```bash
$ ping https://github.com
ping: cannot resolve https://github.com: Unknown host
```


dansholds:
Thanks for the catch!

You're correct that a URL is protocol + QDN such as https://github.com where as a hostname can be just the QDN i.e. github.com

PR #7174, opened 2024-09-20
Update what-is-hosting@aqMaEY8gkKMikiqleV5EP.md

ZlatanCN:
Add new article for 'Web Hosting', it has clearly explained all the details for the beginners. And it will be more reasonable to have not only videos but also article


dansholds:
LGTM thanks! 😄

PR #7211, opened 2024-09-23
changed example to use toUpperCase() instead of .at(0) because .at(0)…

psaradhi:
Updated to use toUpperCase instead of at(0) because .at(0) does not cause any errors since it is available in string and array


dansholds:
If you're going to do this then you need to alter the code snippet below it also.

dansholds (changes requested):
see comment

PR #7233, opened 2024-09-25
feat: add referral user count

arikchakma:
image


kamranahmedse:
@arikchakma good start, let's do it like following. There should be a way for people to invite more members.
Also let's add the leaderboard for this.

image

PR #7273, opened 2024-09-30
Update 100-basic-syntax.md

vamus092:


dansholds:
Not adding much value & syntax mistakes

Issue #7287, opened 2024-10-01
Improve readability of "Done" and "Skipped" items by reducing line-through thickness


faisalmh4045:
### Detailed Description

I’ve noticed that when an item is marked as "Done" or "Skipped," the line-through applied to the text is a bit thick, making the text harder to read. I want to suggest reducing the `text-decoration-thickness` to improve readability. This adjustment would enhance clarity while still clearly indicating the item's status.

papali1305:
Hello sir , i can solve this issue

developer-mohitsoni:
Hi @faisalmh4045, could you attach an image of the problem you're facing?

papali1305:
yes sir it will be very help full if u attach the image

faisalmh4045:
Hi @developer-mohitsoni, I’ve attached an image below to highlight the issue.
![Group 1 (1)](https://github.com/user-attachments/assets/0c8a287c-118d-4ab1-a0a1-e1faf29d7e41)

As you can see, for the "Done" and "Skipped" topics (highlighted with red arrows), the text is styled with a line-through. However, due to the thickness of the line, the text becomes a bit difficult to read.

developer-mohitsoni:
Hi @faisalmh4045, I've already forked the repository and would like to work on this issue. However, I encountered this message while running the project live:

![Screenshot 2024-10-01 212615](https://github.com/user-attachments/assets/5d2bcc34-46db-4e0a-b5a2-1f21abb9301f)

![Screenshot 2024-10-01 212723](https://github.com/user-attachments/assets/a82d8e3c-ae65-456b-ad54-d7915e30b9a8)

kamranahmedse:
Thank you for your suggestion but this is intentional.

PR #7289, opened 2024-10-01
fix: Improve readability of Done and Skipped

developer-mohitsoni:
Fixes:-  #7287


kamranahmedse:
Thank you but the current implementation is intentional.

Issue #6198, opened 2024-07-20
Readability


edusher:
### Feature Description

It would be nice to see an option to view the explanation of each topic in either full-screen or in at least a larger window than the one that currently only pops out a little bit from the right. This would really help readability and would increase the rate at which I could go through each topic!

dansholds:
This is a perfectly valid request and one that has been requested before, we are a small team with a lot to do before we hit requests like this, but I'll leave this open so we keep it in mind!

noobyco:
I was thinking that maybe we could make this feature possible if there is nothing to finish first. How about making it 80% width and height almost filling the screen with an option to go full screen? A little odd-on to that is to add a shortcut (ESC) to close out of the full screen, one more time (ESC) will close that explanation. Or this shortcut thing can be put into a separate issue all together.

PR #7345, opened 2024-10-05
Fixing Readability#6198

SOUMITRO-SAHA:
### Description:
This PR introduces a new feature that allows users to toggle between full-screen mode and a minimized view for the topic explanations. Previously, the explanation window would only pop out a small portion from the right, which could affect readability and user experience. 

Issue #6198 

#### Key Changes:
- Added a toggle button to switch between full-screen and minimized view for better user control.
- When in full-screen mode, the explanation content will be displayed in the center, similar to an article reading layout, improving focus and readability.
- Users can now easily maximize the content to full screen for an immersive experience or minimize it back to the side panel as before.
- Added smooth transitions between the full-screen and minimized views for a seamless user experience.

#### Benefits:
- Improves readability by allowing users to read explanations in a larger view or full-screen mode.
- Offers better flexibility in how users want to consume the content based on their preference.

#### Screenshots:
image
image

### DIFF
--- a/src/components/TopicDetail/TopicDetail.tsx
+++ b/src/components/TopicDetail/TopicDetail.tsx
@@ -526,25 +526,43 @@ export function TopicDetail(props: TopicDetailProps) {
 
               {paidResourcesForTopic.length > 0 && (
                 <>
-                  <ResourceListSeparator text="Premium Resources" icon={Star} />
+                  <ResourceListSeparator
+                    text="Premium Resources"
+                    icon={Star}
+                    className={cn(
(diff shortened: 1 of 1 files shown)


SOUMITRO-SAHA:
Demo Video

[Demo Video](https://www.youtube.com/watch?v=CyKfkKbGNs8)

kamranahmedse:
Thank you for the PR but we would like to tackle this ourselves (if needed). Goal of this popup is to just give you a brief overview and provide external learning resources; so people don't spend that much time on this popup.

PR #7346, opened 2024-10-05
Update npm package manager to include npm audit and yarn audit instructions

lirantal:
Add articles about proper security mitigations against vulnerabilities when using the npm package manager (npm audit) in this npm item on the roadmap. Also added an article for yarn audit for developers using yarn.


offensive-vk (commented):
Doesn't seem correct here in Nodejs Roadmap. (So many links already)
What do you think? @arikchakma

lirantal:
It's specifically for the npm package manager which these articles are focused on. Do you see any other place it makes sense to move them to?

arikchakma:
Thank you for the PR, but we don't encourage adding personal resources.

Issue #7429, opened 2024-10-10
Create an Animation for Developer-Roadmap using Python Turtle


VadisettyRahul:
### Detailed Description

Creating an animation that uses Python’s Turtle library to draw and animate the text "Developer-Roadmap." The animation is  display the text with some creative movements, such as drawing each letter with a pause between each.

1. Expected Result:
The Turtle should draw the text "Developer-Roadmap."
Use animations such as drawing letters sequentially.

2. Requirements:
Use the Python turtle library.
The animation should be interactive and visually engaging.

tasbi03:
Can you assign this to me and if I work toward this are you okay to add hacktoberfest tag on it so that it counts towards it ?

PR #7430, opened 2024-10-10
Implemented Developer-Roadmap Animation using Python Turtle

VadisettyRahul:
Added an animation that uses Python Turtle graphics to draw the text 'Developer-Roadmap' with pauses and sequential drawing of each letter.


arikchakma:
Thank you for your time, but please create an issue before implementing such things.

VadisettyRahul:
I think i made a mistake but i open the issue you can check out #7429

PR #7452, opened 2024-10-12
Working with Files by AI.py

VadisettyRahul:
With the help of this we can extract key information from the very messy text in a very little time.


fellalli:
What exactly would be the use case of this in roadmap.sh? Also, where does `helper_functions` come from?

arikchakma:
Thank you for the effort, but please raise an issue before implementing such things.

PR #7458, opened 2024-10-13
Add animations content to frontend roadmap

orpcy:


orpcy:
Can you provide a guide to naming files? I can see the files have random IDs attached to the end, but I can not find any guide on this

dansholds:
@orpcy You cannot create new roadmap nodes by simply creating a new file, it wouldn't be rendered anywhere and would just sit in the repo.

If you want to suggest a new node/topic be added, then you need to open an Issue.

Screenshot 2024-10-14 at 09 56 41

PR #7464, opened 2024-10-13
feat: added content to AI ​​Engineer Roadmap

deniandreawan:
added content to AI ​​Engineer Roadmap:

- What is an AI Engineer?
- AI Engineer vs ML Engineer
- LLMs


dansholds:
@deniandreawan The last two topics are fine but the first one does not follow contribution guidelines, lets try to keep things to a single concise paragraph.

kenshanta:
Comparing ML and AI engineers while not having a "ML engineer" roadmap itself, does not add up. Is it a sort of reference to another title like "MLOps" ?

deniandreawan:
> @deniandreawan The last two topics are fine but the first one does not follow contribution guidelines, lets try to keep things to a single concise paragraph.

Thank you for the feedback, I’ve updated it https://github.com/kamranahmedse/developer-roadmap/pull/7464/commits/87e464701ab62d953c2460d4a32f0dee9d3b6451

dansholds:
> Comparing ML and AI engineers while not having a "ML engineer" roadmap itself, does not add up. Is it a sort of reference to another title like "MLOps" ?

We have an ML Engineer roadmap in development, not like that means we can't reference another role within an existing roadmap.

PR #7578, opened 2024-10-25
Adding Kafka Streams documentation

eshaan-oreo:
Adding Kafka Streams documentation

### DIFF
--- a/public/roadmap-content/backend.json
+++ b/public/roadmap-content/backend.json
@@ -1683,7 +1683,7 @@
   },
   "VoYSis1F1ZfTxMlQlXQKB": {
     "title": "Kafka",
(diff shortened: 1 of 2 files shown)


dansholds (changes requested):
Hi @eshaan-oreo, you do not need to add new content to both the markdown and the JSON, the JSON is updated at build time from the content in the Markdown.

This contribution is not in line with our contribution guide however so lets remove the content that you added and just keep the new links!

eshaan-oreo:
Hey @dansholds. I Have updated the files. 
I hope you can merge it now 

I plan to write a new road map for Kafka and Kafka Streams, as I am a developer using them. 
Will appreciate if you can give the hacktober-accepted badge 

Thanks

PR #7587, opened 2024-10-25
Update 102-big-omega.md

HunterG333:
Fixed inaccuracy in description


kenshanta:
Both time and space complexity are evaluated taking the <b>worst case</b> into consideration and not the best.

iammursal:
@kenshanta
Big O: Worst Case
Big Omega (Ω): Best Case
Big Theta (Θ): Average Case

https://www.geeksforgeeks.org/difference-between-big-oh-big-omega-and-big-theta/#:~:text=in%20other%20words-,best%20case,-).%0AJust%20like

PR #7600, opened 2024-10-26
Update readme.md | improved socials, formatting.

Piyush-r-bhaskar:


kamranahmedse:
Thanks, but we are fine with what we have at the moment.

PR #7684, opened 2024-11-05
Fixes typo in 104-proc-priorities.md

xaviduds:
Old: renice +5
New: renice -5

From my research, after reading the topic in the Linux roadmap, it didnt make sense that increasing the priority of a process was made by +5, the topic said that a negative number makes the priority higher, so do many articles on the internet.


xaviduds:
I can also make a change that instead of being a minus five that increases a priority it is a plus five that decreases the priority

kamranahmedse:
Thank you!

PR #7686, opened 2024-11-05
Fixes typo in 100-inodes.md: from handle to handled

xaviduds:
From handle to handled, I believe this is grammatically correct.


xaviduds:
So in the commit on line 14 there seems to be no change but the "```" is marked as deleted and created, I don't know why, is that a problem? Why did that happen? Internet gave me nothing. ps: I made the changes and PR in github.com

fellalli:
> So in the commit on line 14 there seems to be no change but the "```" is marked as deleted and created, I don't know why, is that a problem? Why did that happen? Internet gave me nothing. ps: I made the changes and PR in github.com

No worries! That shouldn’t be a problem :)
Your PR looks good!

arikchakma (approved):

Issue #3300, opened 2023-01-14
[Suggestion] Remove links to GeeksforGeeks articles


nielsenjared:
Hey @kamranahmedse ! I love the roadmaps, but I'm finding a lot of references to Geeks for Geeks articles. I would suggest / recommend replacing them for at least two reasons: 
* They aren't a primary source and the accuracy of their content isn't validated. If there's no definitive source, then I would suggest linking to a Wikipedia article. It may not be 100% accurate, either, but at least it's open. 
* They plagiarize the work of others through blatant and brazen copy/pasta without giving credit where credit is due. Featuring them on roadmaps.sh creates backlinks that help them beat out better and more deserving sources in search. 

I would be more than happy to help find alternatives. Let me know! And keep up the good work!

kamranahmedse:
Thank you for the kind words Jared. Oh, I didn't know about the second point. We will keep that in mind in the future.

I would love some help finding alternatives. Feel free to open a PR by replacing the links.

kamranahmedse:
Hey there!

Looks like this issue has been hanging around for a bit without much action. Our roadmaps have evolved quite a bit since then, and a bunch of older issues aren't really applicable anymore. So, we're tidying things up by closing out the older ones to keep our issue tracker nice and organized for future feedback.

If you still think this problem needs addressing, don't hesitate to reopen the issue. We're here to help!

Thanks a bunch!

PR #7705, opened 2024-11-09
Added a resource for type casting

gavhu10:
https://www.geeksforgeeks.org/type-casting-in-python/


kamranahmedse:
Thank you for the PR but we are [trying to avoid geeksforgeeks articles](https://github.com/kamranahmedse/developer-roadmap/issues/3300).

gavhu10:
That makes sense. Thank you for your time!

PR #7714, opened 2024-11-11
Update local-vs-global-config@tVvxC5JQgUb_B8kOqYpD8.md

RudinMaxim:
This link no longer works, at least I couldn't follow it


fellalli:
The link works for me

RudinMaxim:
> The link works for me

Then this site is simply not available in Russia (

PR #7795, opened 2024-11-26
introduction-to-llms.md - editing a broken link and an image, and punctuation

ultumi:
…ctuation

- Editing the URL of the image, it was broken.
- Editing the URL of BERT, it was broken.
- Some punctuation corrections


ultumi:
The BERT link was broken, but it now works. So, please ignore that change.

arikchakma (approved):

PR #7808, opened 2024-11-28
Fix typo in backend beginner

Hidayathamir:
"Learn about differeng testing techniques and learn learn how to write automated tests."

differeng to different


fellalli:
Thanks! But the JSON is auto-generated and edits will be overwritten automatically.

kamranahmedse:
+1 to what @fellalli said, we will fix it in the upcoming release.

PR #7863, opened 2024-12-10
add HPA link

goenning:
adding an article that explain how HPA works


arikchakma:
Sorry, promoting personal content is not allowed. Thank you for your understanding.

PR #7865, opened 2024-12-10
Fix: Fit footer social media icons in one line

xaviduds:
Adjusting tailwind's left and right margin on Footer.astro middots from 0.375rem to 0.2109rem made all the social media icons (mail, youtube and x/twitter) fit into one line. I chose 0.2109rem because after experimenting with it, it was the highest number that did the job.


kamranahmedse:
It is already in one line, thank you for the PR though.

image

PR #7929, opened 2024-12-23
Fix incorrect descriptions of asymptotic notations

fuchen-shi:
It seems that the current descriptions of the Big O and Big Omega notations reflect a common misconception that they correspond to the worst and best cases of an algorithm. Based on what I've learned from my research, every algorithm can have its best case, worst case, and average case runtimes, and each of these runtimes can be represented by a function. The growth rate of such a function can be constrained by upper and lower bounds, and these bounds are what the Big O and Big Omega notations represent.

This PR rewrites the descriptions of the different asymptotic notations based on the above understanding, while adding a new article that provides a thorough explanation of these concepts. Please feel free to share your views on the changes.


public/roadmap-content/computer-science.json
arikchakma, line 1: Let's not update these content JSON files as they are auto-generated.

fuchen-shi: I've rolled back the changes made to the JSON file.

arikchakma (approved):

PR #7934, opened 2024-12-24
Add link to Medium article about pseudocode in the roadmap section

euvitorti:
I included a link to a [Medium article about pseudocode](https://medium.com/@jvitor.ti03/what-is-pseudocode-and-why-should-you-use-it-6a8f9daf5f44) in the roadmap section. The article explains the concept of pseudocode, its benefits, and how to write it effectively. It also provides a practical example of determining if a number is even or odd using pseudocode.

**Purpose:**

- Offer an additional resource for readers interested in learning about pseudocode.
- Enhance the roadmap by linking to an article that clarifies the value and use of pseudocode in programming.

**Changes Made:**

- Added a link to the Medium article under the pseudocode section of the roadmap.


kamranahmedse:
Thank you for the PR but we don't allow promoting personal content.

PR #7938, opened 2024-12-26
Update Interface definition

FarshidABZ:
Add more detail 
Add sample article


kamranahmedse:
Thank you for the PR, but we are intentionally keeping the content under the topic popup concise. There must be a single paragraph, and the focus should be on providing good external resources for learning the topic.

PR #8010, opened 2025-01-10
Update index.md with video

mrlane51:
Added a useful detailed video of Resource Management that is easily digestible.


mrlane51:
added video for roadmap resource list

PR #8012, opened 2025-01-10
feat: add subscribe to changelog

arikchakma:
## Logged Out User
image

## Logged In User
image


src/components/SubscribeToChangelog.tsx
```diff
@@ -0,0 +1,15 @@
+import { showLoginPopup } from '../lib/popup';
+
+export function SubscribeToChangelog() {
+  return (
```
kamranahmedse, line 4: We shouldn't show this if user is logged in.

kamranahmedse (changes requested):
Please add pictures of how it looks like for logged out and logged in user.

kamranahmedse:
@arikchakma I have updated the design, also changed how we are triggering popup

image

PR #8018, opened 2025-01-11
fix: broken link

viksuper555:
The link is broken. Please update to a short link, if used.


arikchakma:
It works

viksuper555:
Yep, now it does.

PR #8055, opened 2025-01-16
Add /proc file system resource

liweisu0:
I append article about `/proc` from kernel doc in `/proc` section


liweisu0:
Sorry I didn't look up the Contribution guide, I will follow the guide and launch another push.

PR #8134, opened 2025-01-28
Fix EPA definition

L4v:
##
The existing text defines EPA (in the context of game development collision resolution) as "Environmental Protection Agency" or "Event-driven Process chain Architecture" which is incorrect.

EPA (Expanding Polytope Algorithm) is used for finding the penetration vector between two convex shapes in order to resolve that collision. I've updated the text and added popular resources to reflect so.

Please let me know of any suggestions!


offensive-vk (commented):
Good work!

PR #8207, opened 2025-02-15
fix(content): ensure consistency in JSON titles

ladunjexa:
Updated "SQL_Select" to "SQL Select" for consistency with other titles.


kamranahmedse:
This is an auto-generated file and will not have any impact on the roadmap content. Any changes should be made in the `src/data` directory.

PR #8217, opened 2025-02-16
Fix small typos

eqsdxr:


kamranahmedse:
This will not have any impact, it's an autogenerated file. You should fix this in the [content directory of roadmap content](https://github.com/kamranahmedse/developer-roadmap/tree/master/src/data/roadmaps/python).

PR #8249, opened 2025-02-20
Update roles-and-responsiblities@K9EiuFgPBFgeRxY4wxAmb.md

havlli:
Second additional resource links to the article that titles AI Engineer, but content actually describes role that is more similar to AI Researcher / ML Engineer


ekrishnachaitanya2004 (commented):
RE

PR #8251, opened 2025-02-20
Do not use `std::endl`

vil02:
[Get rid of std::endl](https://github.com/cpp-best-practices/cppbestpractices/blob/master/08-Considering_Performance.md#get-rid-of-stdendl).


vil02:
@kamranahmedse: it was fixed partially: the ecbb625a093a44f204095cb7852b342df11dfe07 part is missing in the fix.

PR #8252, opened 2025-02-20
`std` is not an abbreviation for STL

vil02:
The current statement is not true. STL is part of the `std` namespace or of the standard library.


kamranahmedse:
Thank you for the PR, can you please resolve conflicts and reopen.

PR #8274, opened 2025-02-24
Use `black` compliant formatting

vil02:
I ran `black` on all python code examples. This is the result.

Similar to #8192.


fellalli:
Good idea to be conform with PEP 8 👍

kamranahmedse:
Can you please resolve the conflicts and reopen the PR

PR #8309, opened 2025-03-08
Update qa.json - delete not actual article

OranciucIvan:
Not actual article for Zombie.js was removed


arikchakma:
Please avoid modifying auto-generated JSON content, instead modify the resource content file. Thank you.

PR #8365, opened 2025-03-21
Replace broken resource URL with a better FreeCodeCamp article

NamanAcharya-Dev:
Replace broken resource URL with a better FreeCodeCamp article  

The previous resource URL was not found, so I replaced it with a more relevant article from FreeCodeCamp:   https://www.freecodecamp.org/news/what-is-an-orm-the-meaning-of-object-relational-mapping-database-tools

### DIFF
--- a/src/data/roadmaps/php/content/object-relational-mapping-orm@SeqGIfcLuveZ2z5ZSXcOd.md
+++ b/src/data/roadmaps/php/content/object-relational-mapping-orm@SeqGIfcLuveZ2z5ZSXcOd.md
@@ -11,4 +11,4 @@ $entityManager->flush();
 
 Visit the following resources to learn more:
 
-- [@article@Object Relational Mapping (ORM)](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/getting-started.html)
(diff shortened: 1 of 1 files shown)


src/data/roadmaps/php/content/object-relational-mapping-orm@SeqGIfcLuveZ2z5ZSXcOd.md
```diff
@@ -11,4 +11,4 @@ $entityManager->flush();
 
 Visit the following resources to learn more:
 
-- [@article@Object Relational Mapping (ORM)](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/getting-started.html)
\ No newline at end of file
+- [@article@What is an Object Relational Mapping (ORM)](https://www.freecodecamp.org/news/what-is-an-orm-the-meaning-of-object-relational-mapping-database-tools)
```
kamranahmedse, line 14:
```suggestion
- [@article@What is an Object Relational Mapping (ORM)](https://stackoverflow.com/questions/1279613/what-is-an-orm-how-does-it-work-and-how-should-i-use-one#answer-1279678)
```

PR #8389, opened 2025-03-24
Update if-else--switch--goto@bjpFWxiCKGz28E-ukhZBp.md

AhmedAlDiab:
I think it now follows the format (one paragraph explaining each topic and a list of resources in the end)
Please provide reasons if it does not.


kamranahmedse:
Please have a look at the contribution docs. The format of your content doesn't match what we require.

PR #8391, opened 2025-03-24
 Corrected description for "Built-in Functions" section

Kouka05:
The previous description primarily explained methods rather than built-in functions, which could cause confusion. This update provides a clearer explanation of built-in functions, including examples like parseInt(), setTimeout(), and Math.random(), while also acknowledging built-in methods within JavaScript objects.

This ensures that the description aligns with the section title and provides more accurate information for learners.


arikchakma (changes requested):
Please have a look at the contribution guideline, we are currently in the process of refactoring the topic resources. Please keep the answer in a paragraph instead of multiple. Thank you so much.

PR #8483, opened 2025-04-09
Added information about Boost. 

DockedFerret800:
Added information about Boost, its advantages, available libraries and how to get started.


DockedFerret800:
@kamranahmedse what's wrong with it?

PR #8506, opened 2025-04-17
fix: handle syntax error

arikchakma:


src/hooks/use-custom-roadmap.ts
```diff
@@ -22,17 +22,25 @@ export function useCustomRoadmap(options: UseCustomRoadmapOptions) {
...
-        }
+          if (secret) {
+            roadmapUrl.searchParams.set('secret', secret);
+          }
+
+          return await httpGet<GetRoadmapResponse>(roadmapUrl.toString());
+        } catch (error) {
+          if (error instanceof SyntaxError) {
+            throw new FetchError(404, 'Roadmap not found');
+          }
```
kamranahmedse, line 40: Put the check on status code instead.

arikchakma: If we wanna do it, we will have to do it inside `httpCall` function.

PR #8516, opened 2025-04-19
Update super-user.md

BryanWV:
Feat: Adding more information to the super user section, and a warning about the power of "sudo"


kamranahmedse:

PR #8529, opened 2025-04-21
Add Java content to the roadmap

vahram-papazyan:
This PR adds various Java-related articles hosted on [JetBrains Academy](https://hyperskill.org/courses) platform by Hyperskill


kamranahmedse:
Thank you for the PR but we don't encourage resources that ask for login/registration. Please feel free to open the PR with different resources.

vahram-papazyan:
> Thank you for the PR but we don't encourage resources that ask for login/registration. Please feel free to open the PR with different resources.

@kamranahmedse I was wondering, did you notice that it asks to register but you can read articles anyway? it is not mandatory

PR #8559, opened 2025-04-26
Add missing content to C++ roadmap

yadunand-kamath:
Added missing content in the _C++_ roadmap: 
- Control flow & statements
- Raw pointers
- Smart pointers
- New/Delete operators


kamranahmedse:
We are not accepting geeks for geeks links, also please look at the contribution docs for content -- it has to be a single paragraph explanation and external links.

PR #8607, opened 2025-05-06
Update Vue composition-api.md

TheJaredWilcurt:
Like the Options-API PR, this is updating the description to be less biased, and focus on the specific benefits of the API.

### DIFF
--- a/src/data/roadmaps/vue/content/composition-api@a0qwdQTunxEjQ9A5wpF-q.md
+++ b/src/data/roadmaps/vue/content/composition-api@a0qwdQTunxEjQ9A5wpF-q.md
@@ -1,6 +1,6 @@
 # Composition API
 
-With the release of Vue 3, developers now have access to the Composition API, a new way to write Vue components. This API allows features to be grouped together logically, rather than having to organize your single-file components by function. Using the Composition API can lead to more readable code, and gives the developer more flexibility when developing their applications.
+Vue offers many approaches for how to write components, including the "Composition API", which is most commonly used via "Script Setup". This approach is based around pulling in low level atomic functions used by Vue's reactivity engine. By composing these low level functions together, you can craft your own system for writing components. One feature this allows for is extending reactivity outside of components. This means you can extract your custom functions for reuse in multiple components. When reusing reactive logic, your functions are called "composables". 
 
 Visit the following resources to learn more:
 


TheJaredWilcurt:
Other notes:

* Might be good to have the 3 main approaches:
  * Options API (Commonly used)
  * Composition API (used by basically no one)
  * Script Setup (Commonly used)
* Currently we are grouping the last two together and calling it "Composition API", even though these approaches are unique enough to be considered separate.

Here is the same component written with all 3 approaches.

```html
<!-- Options API -->
<script>
const colorMessageMap = {
  gold: 'You did it!',
  silver: 'Great job!',
  bronze: 'Well done'
};
export default {
  name: 'MedalComponent',
  props: {
    color: {
      type: String,
      validator: function (value) {
        return Object.keys(colorMessageMap).includes(value);
      }
    }
  },
  computed: {
    message: function () {
      return colorMessageMap[this.color];
    }
  }
};
</script>
```

```html
<!-- Composition API -->
<script>
import { computed } from 'vue';
const colorMessageMap = {
  gold: 'You did it!',
  silver: 'Great job!',
  bronze: 'Well done'
};
export default {
  name: 'MedalComponent',
  props: {
    color: {
      type: String,
      validator: function (value) {
        return Object.keys(colorMessageMap).includes(value);
      }
    }
  },
  setup: function (props) {
    const message = computed(() => {
      return colorMessageMap[props.color];
    });
    return {
      message
    };
  }
};
</script>
```

```html
<!-- Script Setup -->
<script setup>
import { computed } from 'vue';

defineOptions({ name: 'MedalComponent' });
defineProps({
  color: {
    type: String,
    validator: function (value) {
      const ALLOWED_COLORS = [
        'gold',
        'silver',
        'bronze'
      ];
      return ALLOWED_COLORS.includes(value);
    }
  }
});
const message = computed(() => {
  const colorMessageMap = {
    gold: 'You did it!',
    silver: 'Great job!',
    bronze: 'Well done'
  };
  return colorMessageMap[props.color];
});
</script>
```

I've specified in my PR's that "Vue offers many approaches for how to write components". These are the 3 official ways that are built in to Vue 3, however there are *many many* other ways. Including plugins for things like a JS Class based approach. Though no one really uses it, so not really worth mentioning specifically. Just explaining why I've started the descriptions for these APIs off this way.

Issue #8617, opened 2025-05-09
System Design: Topic Not Found


marciopivato:
### Roadmap URL

https://roadmap.sh/system-design

### Suggestions

System Design > Cloud Design Patterns > Messaging > Queu-based Load Leveling  ****** Topic Not Found (missing a d letter)

kenshanta:
@marciopivato will you handle it?

PR #8618, opened 2025-05-09
Fix System Design: Queue based load leveling

AbhishekSrikanth:
- Fix typo in file and topic name.
- Add extension (.md) to file name.
- Relates to #8617


fivemtebex:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #8621, opened 2025-05-10
Update index.md

BryanWV:
As apt is gaining more traction, specially for end users, is better to maintain this command in the summary for Debian Based Distributions


offensive-vk (commented):
Good work! Bryan, but try to make all changes relating to a single roadmap in a single pr where possible and thoroughly review the contributing guidelines to make sure you follow the required style guide.

P.S.: Try to keep the topic content short and easy to understand. Let the links do the talking.

cc #8623 #8622

PR #8635, opened 2025-05-13
docs: fix moved link Managed Resources --> Managing Workloads

serut:
I spent some time to be sure that the link on the card `Resource Management` on roadmap website was pointing to that kubernetes doc https://kubernetes.io/docs/concepts/workloads/management/ and yes it is. 

Kubernetes website MR is https://github.com/kubernetes/website/pull/45212

I don't understand why the card `Resource Management` uses that guide, but if you do, we can at least use the newest page title. In my opinion, we can remove the link instead of accepting this MR, as it's not related to `Resource management`

Hope this MR will help to make roadmap.sh a little better than what it is already <3


nilbuild:
Thank you for the PR, we have made changes to the roadmap which makes this PR outdated.

PR #8643, opened 2025-05-14
doc: note to make it more clear about the confusion

hidaytrahman:
The original description is great; I suggested a note  to make it clear with the JS side of the bug as null is considered an object if we do typeof null.


hidaytrahman:
Hi @kamranahmedse  Can you please have a look on this PR and merge?

nilbuild:

PR #8732, opened 2025-06-01
Update comparison-operators@-v4_V4UuoZSgUk2sqOCim.md

secureman:
Same article referenced twice!!


nilbuild:
Thanks for removing the duplicate link! For future PRs, please use a descriptive title that explains what you're changing (like 'Remove duplicate comparison operators link') instead of just the filename. This helps everyone understand your changes better 👍

secureman:
Okay thanks. Sorry about that will do next time.

PR #8747, opened 2025-06-06
docs: remove unavailable YouTube course link

dansenpir:
The mentioned course is no longer accessible, so it was removed from the roadmap to avoid confusion.


danpamine:
After trying access again, it occurred without failure, then pull request closed.

PR #8779, opened 2025-06-12
Updated iOS Roadmap comment on using static libraries

PyshnovT:
**What happened?**
Removed misleading comments about increased app size when using static libraries.

**Why?**
Static libraries can be stripped of unused symbols during linking, so sometimes they actually have _less_ size than dynamic libraries. The main concern when using static libraries is when you have app extensions and the library is duplicated in both your extension AND your app, thus increasing your app's .ipa size.

**More:**
https://www.emergetools.com/blog/posts/static-vs-dynamic-frameworks-ios-discussion-chat-gpt#static-frameworks-and-size
https://www.emergetools.com/blog/posts/make-your-ios-app-smaller-with-dynamic-frameworks#making-a-dynamic-framework-with-swiftpm


Boxofficefan:
Mm

PR #8811, opened 2025-06-20
refactor(linux): cleanup roadmap content

offensive-vk:
*cleanup and organise content*


nilbuild:
Thanks @offensive-vk, if you are up for another one then I would recommend [rust roadmap](https://roadmap.sh/rust) :)

PS, their rust book is quite good so we can add links from that in most of the nodes

PR #8926, opened 2025-07-17
Add Angular Roadmap Content

jhuang007007:
add interactive tutorial link from official angular documentation to Component Anatomy node


kamranahmedse:
Thanks for adding an official Angular resource! However, the link appears to redirect to the Angular homepage rather than the specific component tutorial. Could you please verify the URL is correct? 

Also, please ensure the file ends with a newline character for consistency with other files.

PR #9024, opened 2025-08-14
fix: update resource label

architxkumar:
The [Golangdocs](https://golangdocs.com/) is not an official link, that's why the label should be updated to `@article`.


nilbuild:
This is a correct fix - golangdocs.com is not official Go documentation, so @article@ is more accurate than @official@. However, this PR has merge conflicts and needs to be rebased. Please rebase your branch and we can merge it.

PR #9092, opened 2025-08-31
Add resource for generics

SeliseRaihanur:


nilbuild:
Personal promotion is not allowed. Thanks for the PR though.

Issue #8940, opened 2025-07-22
Intermediate Level and Advanced Level Buttons dont work


JilvinAbraham:
### What is the URL where the issue is happening

https://roadmap.sh/questions/backend

### What browsers are you seeing the problem on?

Microsoft Edge

### Describe the Bug

The Intermediate and Advanced Level buttons do not take you to the relevant sections of the page

<img width="323" height="277" alt="Image" src="https://github.com/user-attachments/assets/7e913195-cfb7-4f50-8421-56c19027ffca" />

### Participation

- [ ] I am willing to submit a pull request for this issue.

Gireeshs02:
I can fix it!

jcanalesluna:
Hi @JilvinAbraham, thanks for your comments. We will fix the issue asap.

resul-elezi:
Could someone please review my PR for this issue?

PR #9123, opened 2025-09-08
Fixes issue #8940

resul-elezi:
**This pull request makes the following changes:**

*Fixes issue #8940  
- Updates the Table of Contents slugs for "Intermediate Level" and "Advanced Level"  
  so they match the existing heading IDs (`id="intermediate"` and `id="advanced"`).  

This ensures that the links correctly scroll to the intended sections.

Note: "Beginner Level" already works both with `beginner` and `beginner-level`, 
because of the way `slugify("Beginners")` is handled.


nilbuild:
Thank you!

PR #9185, opened 2025-09-26
Fix: Updated navigation logic and URL fix for AI courses

FenchsApps:


nilbuild:
Thank you for the PR, however we are currently revamping the AI tutor and this will be improved.

PR #9202, opened 2025-09-30
Add link to file a stumped page

TazyFoundSoup:
This adds a new link to the page for people to report a stump page on GitHub. It's just in case something goes wrong.


TazyFoundSoup:
And sorry if it may be buggy, I'm away from home so I don't have my astro installation so I'm doing everything from Codespaces :]

nilbuild:
Thanks for the PR! I don’t think we need this though; 404s are pretty rare and usually not something users need to report. Let's keep things simple for now.

PR #9220, opened 2025-10-05
enhance performance phase 1 

abdo-spices:
### before:
image

### after: 
image


nilbuild:
Thanks for the PR! We’re in the middle of a full refactor of this area, so the patch will be obsolete shortly. Rather than putting it through review and testing now, we'll fold the ideas into the new implementation.

PR #9289, opened 2025-10-24
Add video link for Context vs. Prompt Engineering

Dasher83:
Added a video link to enhance understanding of role prompting.


jcanalesluna:
Thanks for submitting the PR. The video you provided contained a tracking URL that we don't allow. Can you remove it before we move on with the PR?

PR #9291, opened 2025-10-24
Add video link to Chain of Thought Prompting

Dasher83:
Added a video link related to Context Engineering and Prompt Engineering.


jcanalesluna:
Thanks for submitting the PR. The video you provided contained a tracking URL that we don't allow. Can you remove it before we move on with the PR?

PR #9317, opened 2025-11-01
fix: ajust multiple mutation fields description

gustavosett:
# Multiple Mutation Fields are **NOT** atomic!

It is crucial to understand that these operations are not atomic by default. According to the specification, mutation fields within a single request are executed serially (one after the other). If one mutation fails, it does not roll back any previously successful mutations in the same request. This means your application must be prepared to handle partially successful data changes.

### **If the first mutation succeeds but the second mutation fails, the first one will not be rolled back.**


gustavosett:
#9318 reference

jcanalesluna:
Thanks for reporting and submitting the PR for clarification :)

Issue #9281, opened 2025-10-22
Typo in error message for 'name' field in Signup API response


sathwikhbhat:
### What is the URL where the issue is happening

https://roadmap.sh/signup

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge, Other, Safari, Firefox

### Description:

When using the Signup API endpoint (`https://roadmap.sh/api/v1-register`), if there is any issue related to the `name` field (for example, when the name exceeds the allowed length), the error message returned contains a typo.

Instead of saying **"Name must be 50 characters or less"**, it incorrectly says **"Mame must be 50 characters or less"**.

### Example Request Body:

```json
{
    "email": "test@example.com",
    "name": "thisisaverylongnamebeyondtheallowedlimitandshouldtriggeranerror",
    "password": "password123"
}
```

### Example Response:

```json
{
    "status": 400,
    "message": "Mame must be 50 characters or less",
    "errors": [
        {
            "message": "Mame must be 50 characters or less",
            "location": "name"
        }
    ]
}
```
![Image](https://github.com/user-attachments/assets/28979beb-41b7-4710-964e-8b16fe7eddc0)

### Expected Behaviour:

The response message should correctly refer to **"Name"** instead of **"Mame"**.

### Expected Correct Response:

```json
{
    "status": 400,
    "message": "Name must be 50 characters or less",
    "errors": [
        {
            "message": "Name must be 50 characters or less",
            "location": "name"
        }
    ]
}
```

### Note:
This issue cannot be fixed directly in the frontend code because the typo originates from the backend API response. It should ideally be corrected in the backend validation message.

However, as a temporary workaround, the frontend could replace all instances of "Mame" with "Name" in any displayed error messages.

### Participation

- [ ] I am willing to submit a pull request for this issue.

jcanalesluna:
This should be fixed now. Thanks for reporting :)

PR #9344, opened 2025-11-08
Added frontend fix for signup name max length validation typo

ronodhirSoumik:
This is a frontend fix (should be temporary - may need to be handled from backend) of 'Name' typo in Signup Api response #9281


osama-mhmd:
Pasted image

Actually, It's "Name". Maybe someone has fixed it. Thx btw for the solution provided

ronodhirSoumik:
Closing PR as the issue is already fixed

PR #9347, opened 2025-11-09
Create web-apis.md with overview and resources

osama-mhmd:
Added introductory content about Web APIs and resources for further learning.


jcanalesluna:
Thanks for submitting the PR. However, we don't accept PRs that involve adding or deleting nodes in existing roadmaps. In this case, please submit an issue with your suggestion, and we will process it. Best

PR #9349, opened 2025-11-10
fix : remove a link to a non existing page

MohamadObeid9:
removed a link to a non existing page that was supposdely leading an official documenatation for go doc


ebcq:
https://go.dev/blog/godoc

PR #9377, opened 2025-11-20
Enhance TRIZ content with detailed problem-solving steps

zemgeez:
Expanded the TRIZ section with detailed steps for problem-solving, including defining the problem, identifying contradictions, using the contradiction matrix, applying inventive principles, and considering ideal final results.


jcanalesluna:
Thanks for reporting and submitting the PR. However, your proposal doesn't meet our current editing style. We're trying to keep descriptions short (eg. 1 paragraph) so we don't overwhelm learners. For additional explanation, learners can check the resources or start a chat with our AI Tutor.

PR #9380, opened 2025-11-21
Add installation guide for Swift on multiple OS

Ramc26:
Added detailed installation instructions for Swift on macOS, Windows, and Linux.


jcanalesluna:
Hi, thanks for submitting the PR. There was an issue with the Swift roadmap, and we lost all the content. This is fixed now, and we should be content with the Installation node. I close your PR now because we want to keep the explanation short (e.g. 1 paragraph) so as not to overwhelm users. For additional explanation, the list of resources should work. In addition, now it's possible to start a chat with the AI tutor to get extra info

PR #9381, opened 2025-11-21
Enhance 'What is Swift?' content with details

Ramc26:
Added a detailed description of Swift programming language.


jcanalesluna:
We had an issue with the Swift roadmap in the last days. It should be fixed now and you should have content in every node of the roadmap. No need to start from scratch. But thanks anyway

PR #9460, opened 2025-12-14
fix: correct MDN typo and update whitespace link

Efimish:
- Fixed typo in link label (`DMN` → `MDN`)
- Updated MDN link to the current canonical URL (avoids 301 redirect)


Efimish:
I found another occurrence of the same typo (DMN) and added a fix in a follow-up commit to this PR.

PR #9515, opened 2026-01-04
fixing redis typo -> added "Setting" word

Rajeshsaharan:
Settings word is added to node


jcanalesluna:
Thanks for the feedback! I close the PR now because node changes are only possible through backoffice. I've fixed the type already, but it may take some time to be displayed on the page due to the cache policy.

PR #9517, opened 2026-01-05
fixing-typo for Prefect in data_engineering roadmap

akshaytayade:
fixed the typo for the node containing details of Prefect which is written as Perfect.

change in: developer-roadmap\src\data\roadmaps\data-engineer\data-engineer.json

Attaching screenshot for reference:
image


jcanalesluna:
Thanks for the feedback! I close the PR now because node changes are only possible through backoffice. I've fixed the type already, but it may take some time to be displayed on the page due to the cache policy.

PR #9568, opened 2026-01-27
fix: typo "Databsases" in data-engineer.json

gonzariosm:
Fix typo Databsases to Databases


jcanalesluna:
Made the change from backoffice, as it's not possible through a PR. Thanks for reporting!

PR #9574, opened 2026-02-02
fix: typo "Eurkea" in Spring Boot roadmap

MrMatiz2:


jcanalesluna:
Hi! I made the change already from backoffice, as node titles cannot be changed using a standard PR. Thanks for reporting :)

PR #9725, opened 2026-03-14
Fix mobile roadmap scaling and navigation with contained pinch-zoom and pan

### LINKED ISSUE #9724: Roadmap scaling and navigation UI/UX is broken on mobile
### Detailed Description

> [!NOTE]
> This issue is resolved in PR #9725.

## Problem Description
On mobile devices with smaller screens, each roadmap is rendered in a desktop-oriented layout, which causes the nodes to shrink to a size that is extremely difficult to read and tap reliably. In practice, the only workable way to navigate the roadmap is to use browser-level pinch-zoom and drag-to-pan.

That creates an immediate problem for the primary interaction of opening a node modal. After zooming in to find and tap a node, the modal also appears zoomed in because the entire page has been scaled. This forces the user to zoom back out just to read the modal content.

When the modal is closed, the roadmap is now zoomed back out as well, so the user loses their place and has to zoom in again to continue navigating. The same problem also affects the rest of the page content, since browser-level zoom is not isolated to the roadmap.

This repeated cycle of zooming in to navigate the roadmap and zooming back out to read the node modal and other page content creates significant friction and makes the mobile experience much more disruptive than the intended desktop experience.

## Video of Issue

https://github.com/user-attachments/assets/71cc0318-82b2-4bd0-b362-c067e33d4c75

ritorhymes:
## Summary

This PR adds contained pinch-zoom and pan to the roadmap on touch devices so users no longer have to rely on browser-level zoom to navigate it. Closes #9724 

It does not change the desktop roadmap experience for mouse and keyboard users, and it does not introduce any new dependencies. The feature is gated behind touch-device detection and is intended for smaller, touch-first screens where the default roadmap scale makes nodes difficult to read and tap reliably.

**Live Demo:** [https://roadmap.ritovision.com/ai-engineer](https://roadmap.ritovision.com/ai-engineer)
*Note: the roadmap node renderer in this demo is visually disorganized because the upstream renderer is not publicly available and a quick makeshift fallback was used locally. Focus on the interaction behavior rather than the rendering fidelity.*

## Problem Recap
From #9724:
Roadmap nodes are too small to read and tap on mobile-sized touch screens at the default page scale. In practice, that leaves browser-level zoom as the only workable way to use the roadmap, but that zoom applies to the entire page rather than to the roadmap itself.

That creates a constant mismatch: users need to zoom in to work with the roadmap, then zoom back out whenever they open a node's modal to read it, then zoom back in again to continue navigating the roadmap while also losing their place from having just zoomed out.

This makes roadmap navigation on mobile much more disruptive than the intended desktop interaction.

## What This Adds

- Add a touch-only roadmap viewport that supports pinch-zoom and drag-to-pan on mobile devices while leaving desktop mouse behavior unchanged. This keeps zoom interaction contained to the roadmap instead of relying on browser-level zoom that affects the entire page and must be repeatedly undone to read node modals and continue navigating.

- Clip overflow to keep the zoomed roadmap contained within its frame and prevent panning from spilling the roadmap outside its container and overlapping other page content.

- Cap the viewport height on touch devices to avoid a scroll-lock situation where the roadmap occupies most of the screen and intercepts touch input, leaving the user with almost nowhere to touch the page and scroll normally. The side margins around the roadmap are too narrow to rely on comfortably, so reserving space above and below it via the viewport height cap preserves a practical area for normal page navigation.

- Show a hint explaining how to navigate the roadmap before zoom is activated.

- Provide a floating reset action that restores the roadmap zoom level to the default state

- Suppress accidental clicks after gesture movement.

- Drag-to-pan is gated until the first pinch-zoom to prevent accidental triggering when scrolling the page.

- Implement the interaction at the roadmap viewport wrapper layer so it does not depend on renderer-specific behavior.

### Default State
*Notice the roadmap does not span the full viewport height to reserve touch areas for the user to scroll vertically around. Also notice the centered floating UI instructions prompting the user to pinch to begin navigating the roadmap. These will NOT render on desktop like this; the roadmap will span the full natural height and not display the instructional UI.*
after-zoomed-out

<br /><br />

### Contained Zoom In
*Notice the roadmap is zoomed in but the rest of the page remains its normal scale, also notice the "Reset Zoom" button that floats at the top of the roadmap container.*
after-zoomed-in

## Testing

I tested the feature locally on Chrome, Firefox and Edge mobile browsers on Android and the behaviors were stable and consistent for each.

I also tested it on Chrome, Firefox and Edge desktop browsers with mouse and keyboard to ensure no regressions emerged for the desktop experience and found no issues.

Live Demo: [https://roadmap.ritovision.com/ai-engineer](https://roadmap.ritovision.com/ai-engineer)
*Note: the roadmap node renderer in this demo is visually disorganized because the upstream renderer is not publicly available and a quick makeshift fallback was used locally. Focus on the interaction behavior rather than the rendering fidelity.*

## Note on Validation

This feature was developed and tested locally in a setup where the roadmap renderer had to be swapped out because collaborator-only renderer access was not available.

Because the interaction is implemented at the roadmap viewport wrapper layer rather than inside renderer internals, it should carry over cleanly. It should still be verified against the upstream renderer on real touch devices.

### Default State
*Notice the roadmap does not span the full viewport height to reserve touch areas for the user to scroll vertically around. Also notice the centered floating UI instructions prompting the user to pinch to begin navigating the roadmap. These will NOT render on desktop like this; the roadmap will span the full natural height and not display the instructional UI.*

after-zoomed-out

<br /><br />

### Contained Zoom In
*Notice the roadmap is zoomed in but the rest of the page remains its normal scale, also notice the "Reset Zoom" button that floats at the top of the roadmap container.*
after-zoomed-in

### Before

https://github.com/user-attachments/assets/71cc0318-82b2-4bd0-b362-c067e33d4c75

### After
***Note**  the roadmap node renderer in this demo is visually disorganized because the upstream renderer is not publicly available and a quick makeshift fallback was used locally. Focus on the seamlessness of zooming in and panning around the roadmap without ever needing to reverse the zoom to read the modal content or the rest of the page.*

[roadmap-after.webm](https://github.com/user-attachments/assets/7d63d984-343d-45e0-85f8-b80dbf077f6d)

### DIFF
--- a/src/components/EditorRoadmap/EditorRoadmap.tsx
+++ b/src/components/EditorRoadmap/EditorRoadmap.tsx
@@ -1,4 +1,5 @@
-import { useEffect, useState, type CSSProperties } from 'react';
+import { useEffect, useRef, useState, type CSSProperties } from 'react';
+import { RefreshCcw } from 'lucide-react';
 import {
   EditorRoadmapRenderer,
   type RoadmapRendererProps,
@@ -11,6 +12,11 @@ import {
 import { httpGet } from '../../lib/http';
 import { getUrlParams } from '../../lib/browser.ts';
 import { RoadmapFloatingChat } from '../FrameRenderer/RoadmapFloatingChat.tsx';
+import {
+  isTouchInteractionDevice,
+  TOUCH_INTERACTION_MEDIA_QUERY,
+} from '../../lib/touch-interaction.ts';
+import { useTouchPanZoom } from './useTouchPanZoom.ts';
 
 type EditorRoadmapProps = {
   resourceId: string;
@@ -32,9 +38,26 @@ export function EditorRoadmap(props: EditorRoadmapProps) {
 
   const [hasSwitchedRoadmap, setHasSwitchedRoadmap] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
+  const [isTouchInteractionMode, setIsTouchInteractionMode] = useState(false);
+  const [isTouchHintVisible, setIsTouchHintVisible] = useState(false);
+  const [isTouchHintFading, setIsTouchHintFading] = useState(false);
   const [roadmapData, setRoadmapData] = useState<
     Omit<RoadmapRendererProps, 'resourceId'> | undefined
   >(undefined);
+  const viewportRef = useRef<HTMLDivElement>(null);
+  const contentRef = useRef<HTMLDivElement>(null);
+  const touchHintFadeTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(
+    null,
+  );
+
+  const clearTouchHintFadeTimeout = () => {
+    if (touchHintFadeTimeoutRef.current === null) {
+      return;
+    }
+
+    clearTimeout(touchHintFadeTimeoutRef.current);
+    touchHintFadeTimeoutRef.current = null;
+  };
 
   const loadRoadmapData = async () => {
     setIsLoading(true);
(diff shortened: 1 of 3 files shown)


jcanalesluna:
Thanks for pointing out this issue. We will put it in the backlog and handle it internally when the time comes.

ritorhymes:
Thanks for the response.

Most of my time spent using roadmap.sh has been on mobile; I'm glad you're aware of the issue now but it's too bad it's being backlogged.

I believe the implementation I've offered here is a strong, contained solution, and it works well right now.

If you revisit it later and want me to pick it back up, refine it, or help validate an approach, feel free to reopen this or reach out.

Rito

PR #9775, opened 2026-03-24
Fix typo of LSTM label on image and markdown

AruchaK:
The LSTM (Long Short-Term Memory) on the website is mislabeled as LSMT. (https://roadmap.sh/machine-learning)
I tried to edit the image to be as accurate as possible to the original image.

image


TrymJohnsen:
good spot

jcanalesluna:
Thanks for reporting. I've fixed this from backoffice, as it cannot be done through a PR!

PR #9854, opened 2026-04-14
docs: add prompt tuning content

Stewie-pixel:
add prompt tuning content for better prompt engineering technique


jcanalesluna:
Thanks for this PR. We cannot create new nodes from a PR. I've done it myself from backoffice. It mimics your content. Thanks!

PR #9872, opened 2026-04-22
docs: add hero image to README

livlign:
## Summary

Adds a hero image to the top of `readme.md` that surfaces the catalog at a glance — brand, tagline, a sample of the tracks, and the stat that you've shipped **85 roadmaps** since 2017.

All on-screen content is pulled from the repo itself:

- **85 roadmaps** — counted from `src/data/roadmaps/` (directory listing)
- **Since 2017** — first commit `2017-03-15` ("Initial commit")
- **Node cloud** — 51 of 85 track names rendered as chips, with an explicit bottom fade + `85 Roadmaps · pick your path` label so the partial listing doesn't misrepresent scope
- **No volatile stats** — no star count, no release number, no download number (these would date the image immediately)

## Changes

- Add `public/img/hero.png` (1200×675 @2x retina, ~135 KB)
- Insert a single `![…]` line in `readme.md` between the centered brand block and the existing wavy `waxVImv.png` separator

Alt text includes the full positioning + stats so screen-reader and text-only viewers get the same information.

## Test plan

- [ ] Open `readme.md` on GitHub and confirm the hero renders between the brand header and the wavy separator
- [ ] Easy revert: delete `public/img/hero.png` + remove the one added line

Feel free to close if the direction isn't aligned — happy to iterate on palette, copy, or composition, or drop it entirely.


nilbuild:
Thanks but no need for that

PR #9874, opened 2026-04-23
add resources to frontend roadmap

maiwand-atssar:
add resources to frontend roadmap git and version control node


jcanalesluna:
Hi, thanks for the PR. We're trying to get no more than 3-4 resources for each node!

Issue #10149, opened 2026-07-13
Rename Gemma2 node to Gemma in AI Engineer roadmap


DorukTaneli:
## Suggested change

Rename the `Gemma2` node in the AI Engineer roadmap to `Gemma`.

## Reason

The other model nodes use model-family names without version numbers. `Gemma2` is the only version-specific label, while `Gemma` represents the current model family and avoids tying the roadmap to an outdated release.

The related content update will also rename the sidebar title and remove the outdated "Gemma explained: What's new in Gemma 2" resource.

jcanalesluna:
Done! Thanks for the suggestion

PR #10150, opened 2026-07-13
Rename Gemma2 topic to Gemma

DorukTaneli:
## What changed

- Renamed the Gemma topic content and sidebar title from `Gemma2` to `Gemma`.
- Updated the description to refer to the Gemma model family.
- Removed the outdated "Gemma explained: What's new in Gemma 2" free resource.

## Why

The other model entries use model-family names without version numbers. `Gemma2` was the only version-specific label, which made the roadmap inconsistent and tied the topic to an outdated release.

The roadmap node label is managed separately from topic content, so its required title-change request is tracked in #10149.

## Impact

The Gemma sidebar content uses the current family name and no longer recommends the outdated Gemma 2 article.

## Validation

- `git diff --cached --check`
- Searched the AI Engineer roadmap content for remaining `Gemma2`, `Gemma 2`, and removed-resource references.


jcanalesluna:
I've just changed the name of the node —this is something that can only be done through the backoffice. If you want to change the description and resources associated with the node, please wait some hours so the cache policy updates the node url. Then you can go ahead!

DorukTaneli:
Cache updated, PR ready for review!

PR #10171, opened 2026-07-22
Update EC2 instance type from t2.micro to t3.micro

SzilvasiPeter:
The t2.micro is not eligible for Free Tier.

See the command output message:

```sh
$ aws ec2 run-instances --image-id ami-0ac1f955d6e62f3f1 --instance-type t2.micro

aws: [ERROR]: An error occurred (InvalidParameterCombination) when calling the RunInstances operation: The specified instance type is not eligible for Free Tier. For a list of Free Tier instance types, run 'describe-instance-types' with the filter 'free-tier-eligible=true'.
```


jcanalesluna:
Thanks for reporting!

PR #10206, opened 2026-08-05
Clarify LLM capabilities and external tool dependencies

mohakhan03uk:
Added a note clarifying that LLMs are next-token prediction models and that additional features depend on external tools.


jcanalesluna:
Thanks for the suggestion. I've modified the description to include your idea, but shortened it to stick to our length formatting

